PR #247, opened 2014-09-09
Update hashes.txt

erosman:
Added Hash list of JQuery UI from jqueryui.com


magopian:
What's the status on this PR? Is it going to be updated?

PR #259, opened 2014-11-18
Bug 1000550 - Make warnings less stern, where applicable

kmaglione:
CC @jvillalobos

### DIFF
--- a/validator/constants.py
+++ b/validator/constants.py
@@ -48,7 +48,7 @@ SPIDERMONKEY_INSTALLATION = os.environ.get("SPIDERMONKEY_INSTALLATION")
 
 DEFAULT_WEBAPP_MRKT_URLS = ["https://marketplace.mozilla.org",
                             "https://marketplace-dev.allizom.org"]
-BUGZILLA_BUG = "https://bugzilla.mozilla.org/show_bug.cgi?id=%d"
+BUGZILLA_BUG = "https://bugzil.la/%d"
 MDN_DOC = "https://developer.mozilla.org/docs/%s"
 
 JETPACK_URI_URL = "https://wiki.mozilla.org/Labs/Jetpack/Release_Notes/" \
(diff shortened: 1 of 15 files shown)


validator/constants.py
```diff
@@ -48,7 +48,7 @@
 
 DEFAULT_WEBAPP_MRKT_URLS = ["https://marketplace.mozilla.org",
                             "https://marketplace-dev.allizom.org"]
-BUGZILLA_BUG = "https://bugzilla.mozilla.org/show_bug.cgi?id=%d"
+BUGZILLA_BUG = "https://bugzil.la/%d"
```
magopian: Why this change?

validator/testcases/javascript/call_definitions.py
```diff
@@ -30,10 +30,9 @@ def amp_rp_bug660359(wrapper, arguments, traverser):
                 "amp_rp_bug660359"),
         notice="Custom add-on types may not work properly in Gecko 6",
         description="This add-on appears to register custom add-on types, "
-                    "which are affected and may not work properly due to "
+                    "which are affected by, and may not work properly due to, "
                     "changes made on Gecko 6. For more information, "
-                    "please refer to "
-                    "https://bugzilla.mozilla.org/show_bug.cgi?id=595848",
+                    "please refer to %s" % BUGZILLA_BUG % 595848,
```
magopian: Nice, much better to factor out the bugzilla url, thanks!

validator/testcases/javascript/traverser.py
```diff
@@ -282,18 +282,26 @@ def _build_global(self, name, entity):
...
+                elif isinstance(dang, dict):
+                    kwargs.update(dang)
+                else:
+                    kwargs["description"] = ("Access to the `%s` property is "
+                                             "deprecated for security or "
+                                             "other reasons." % name)
+
+                self._debug("DANGEROUS")
+                self.err.warning(**kwargs)
+
```
magopian: Much more readable that way, thanks

magopian:
lgtm, should I wait for @jvillalobos to r+ before merging?

jvillalobos:
The changes look good to me.

validator/constants.py
kmaglione: Because the full URLs are visible in the validation messages, and the longer URLs add a lot of noise and distracting line breaks.

magopian: Makes sense, thanks for the explanation

PR #266, opened 2015-01-12
release 1.4.6 to pypi

magopian:
I chose to release 1.4.6, which is the next logical version when you look at
the tags, and not 1.1, which is the next logical version when you look at the
current version on pypi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 sudo: false
 python:
-  - "2.6"
+  - '2.6'
 env:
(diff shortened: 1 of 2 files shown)


magopian:
@mstriemer this needs that you grant the access on pypi to the `marketplacedevsinternal` user, could you please do that? Thanks!

mstriemer:
I gave the user the Owner role. I'm not sure if that's the right thing to do but it seems like the shared account should be an Owner.

mstriemer:
r+

PR #267, opened 2015-01-13
Add deprecation warning for use of the widget API now that it's slated f...

kmaglione:
...or removal.


magopian:
Excellent work, thanks a lot @kmaglione

PR #261, opened 2014-12-02
Issue a warning when add-on is already signed. (bug 1070188)

olivier-m:
I added a test to check if an add-on has a preexisting signature.
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1070188

The test only checks the presence of signature files, not their contents.

PR #268, opened 2015-01-15
Issue a warning when add-on is already signed. (bug 1070188)

magopian:
Fixes [bug 1070188](https://bugzilla.mozilla.org/show_bug.cgi?id=1070188)

### DIFF
--- a/validator/testcases/content.py
+++ b/validator/testcases/content.py
@@ -258,6 +258,23 @@ def test_packed_scripts(err, xpi_package):
             err.pop_state()
 
 
+@decorator.register_test(tier=2)
+def test_signed_xpi(err, xpi_package):
+    """Checks if XPI is signed."""
+    search = set([
+        'META-INF/manifest.mf',
+        'META-INF/zigbert.rsa',
+        'META-INF/zigbert.sf'])
+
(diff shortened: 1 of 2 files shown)


magopian:
@mstriemer @clouserw r?

This supersedes #261

validator/testcases/content.py
```diff
@@ -258,6 +258,23 @@ def test_packed_scripts(err, xpi_package):
             err.pop_state()
 
 
+@decorator.register_test(tier=2)
+def test_signed_xpi(err, xpi_package):
+    """Checks if XPI is signed."""
+    search = set([
+        'META-INF/manifest.mf',
+        'META-INF/zigbert.rsa',
+        'META-INF/zigbert.sf'])
```
mstriemer: Is this the only change?

mstriemer:
Only seeing the one change, lgtm. r+

validator/testcases/content.py
magopian: This line is correctly indented, and the `tier` in the decorator was changed from `1` to `2` on your advice.

magopian:
Thanks for the review!

PR #269, opened 2015-01-28
Gecko 35 compat

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1122229
https://bugzilla.mozilla.org/show_bug.cgi?id=1121744
https://bugzilla.mozilla.org/show_bug.cgi?id=1121732

### DIFF
--- a/validator/testcases/regex.py
+++ b/validator/testcases/regex.py
@@ -1535,6 +1535,16 @@ class Gecko35RegexTests(CompatRegexTestHelper):
             log_function=self.err.warning,
             compat_type="error")
 
+        iterator_link = (MDN_DOC % "Web/JavaScript/Reference/Global_Objects"
+                                   "/String/@@iterator")
+        yield self.get_test(
+            """['"]@@iterator['"]""",
+            "The syntax for @@iterator has changed.",
+            "The syntax for @@iterator has changed. See {link} for more "
+            "information.".format(link=iterator_link),
+            log_function=self.err.warning,
+            compat_type="error")
+
 
 #############################
 #  Thunderbird Regex Tests  #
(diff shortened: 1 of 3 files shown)


validator/testcases/regex.py
```diff
@@ -1516,6 +1516,36 @@ def tests(self):
...
+class Gecko35RegexTests(CompatRegexTestHelper):
+    """Regex tests for Gecko 35 updates."""
+
+    VERSION = FX35_DEFINITION
+
+    def tests(self):
+        box_object_subject = ("All interfaces derived from nsIBoxObject no "
+                              "longer exist.")
+        yield self.get_test(
+            r"\bnsI[\w\d]*BoxObject\b",
```
magopian: why the `\d`?

mstriemer: Probably not necessary. It's what I used to get the list of objects but I don't think anything matches it. I'll remove it.

validator/testcases/regex.py
```diff
@@ -1516,6 +1516,36 @@ def tests(self):
...
+            "QueryInterface it's likely you just need to remove "
+            "the function call. See %s for more information."
+            % BUGZILLA_BUG % 979835,
+            log_function=self.err.warning,
+            compat_type="error")
+
+        iterator_link = (MDN_DOC % "Web/JavaScript/Reference/Global_Objects"
+                                   "/String/@@iterator")
+        yield self.get_test(
+            r"['\"]@@iterator",
```
magopian: Just a suggestion: use `r"""..."""` to avoid having the `\` in there, maybe it's clearer?
Also, you're matching on the first quote, would that make sense to also match on the last one? Or at least the word boundary `\b`?

magopian:
r+wc, nice work!

validator/testcases/regex.py
mstriemer: Ah, yeah I meant to match on both, must have forgotten. I thought about doing a `"""` then didn't because the one `\` was shorter. It's probably a good call though. `\b` doesn't seem to match quotes (or didn't make the test pass anyway).

PR #270, opened 2015-02-03
JS libs whitelist update (no bug)

wagnerand:


magopian:
Thanks you very much @wagnerand !

PR #271, opened 2015-02-04
Refactor framework and library detection code.

kmaglione:
Sorry. This is kind of large.

Thoughts for easier review:

The `JetpackCFX.validate` method is mostly code that was in 'jetpack.js', but with a lot of dead code removed.

The 'test_libraries.py` file is mostly code from the very strangely misnamed 'test_libraryblacklist.py' file, but git's rename tracking is not smart enough to notice.

The code in 'metadata/libfetcher.py' and 'metadata/update.py' can be skipped, if you prefer. It won't be run on servers, and we'll probably be changing it locally without submitting PRs immediately for a lot of runs.

There's a related update to Olympia coming to handle out-of-band updates of the metadata JSON and add some integration with the file viewer.


kmaglione:
Oh, and fortunately most of the "files changed" are just deletions of old cruft and metadata files.

magopian:
This PR is not mergeable anymore, could you please rebase it?

When git isn't able to detect the renaming/moving of files, you can help it, I believe, by using `git mv`. Maybe this would make this PR more readable?

I'm afraid I'm not familiar enough with the codebase (yet?) to review this PR, maybe @mstriemer will have some time to give it a look?

diox:
Bug number ? What is the plan exactly here, if you are removing the old lists of files ?

mstriemer:
I didn't get to look at all the code but the cleanup looks a lot nicer than the old code, thanks.

I'm assuming the out-of-band changes PR in olympia will allow for `libraries.json` to be updated outside of git. Will this update the `libraries.json` file on the server or is there some other mechanism to get those changes into the validator?

How will the out-of-band changes interact with the checked-in version?

I'm just wondering what the game plan is here so I know what to expect in all of these changes.

kmaglione:
Huh. I rebased before pushing. I guess I'll try again.

I did use `git mv`, but unfortunately it doesn't make a difference.

kmaglione:
Related bugs: https://bugzil.la/1069570 https://bugzil.la/1013413

Yes, the out-of-band changes will be to update the libraries.json data. The file itself won't be changed. AMO will pass the data directly to the `validate` method, much like it currently does for the application metadata.

magopian:

kmaglione:
Rebased and fixed some conflicts.

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
+# Directory in which downloaded temporary library files may be
+# stored.
+LIBS_DIR = os.path.join(SCRIPT_DIR, "jslibs/")
+
+# Used to throw away betas, RCs, etc.
+VALID_VERSION = re.compile(r"^\d[\d.]+$")
+
+
+def filter_re(expr):
```
magopian: docstring?

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+    return re.compile(expr).search
+
+
+def tag_transform(tag, min_version=None, prefix=None):
+    """
+    Generic function to transform a revision control tag to a version number.
+    Returns `None` for invalid tags.
```
magopian: Give an example of what that does (eg, with a prefix)

magopian: And add an explanation of what/why the min_version

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+            self.library_id = library_id
+
+    def get_versions(self):
+        """
+        Abstract method which returns an iterable of tuples,
+        containing a version number and an opaque, JSON-compatible
+        identifier for the location of the version in question
+        (usually the URLs of scripts).
+        """
+        return ()
```
magopian: `raise NotImplementedError` ?

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+
+    def get_files(self, version, location):
+        """
+        Abstract method which, given a version and location as
+        produced by `get_versions`, returns an iterable of tuples,
+        which contain the path and contents of the file.
+        """
+        raise NotImplementedError
+
+    def get_checksums(self, version, location):
```
magopian: not used? shouldn't be used?

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+    def get_files(self, version, files):
+        for key, url in files.iteritems():
+            filename = os.path.basename(urlparse(url).path)
+            path = "%s/%s" % (key, filename)
+
+            r = requests.get(url)
+            yield path, r.content
+
+
+class ZipFetcher(object):
```
magopian: Please add `Mixin` to the name: `ZipFetcherMixin`, it helps spotting it ;)

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    Mixin for handling libraries served as zip archives.
+
+    Supports fetchers whose `get_versions` methods return the URL of a zip
+    archive, and which optionally implement a `file_filter` method to restrict
+    the checksummed files.
+
+    The entire file is read into memory.
```
magopian: I believe this line/comment could be a docstring of the `get_files` method

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+
+    def get_files(self, version, url):
+        from StringIO import StringIO
```
magopian: add a comment why it's not using `cStringIO`

metadata/libfetcher.py
```diff
@@ -0,0 +1,373 @@
...
+
+        r = requests.get(url)
```
magopian: no one-letter variables please

tests/resources/bootstrap-jpm.js
```diff
@@ -0,0 +1,385 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
magopian: This is a test framework used in the tests to make sure it's correctly detected as a framework, right? If so, could it be much shorter, maybe even just containing a line saying that it's a test framework?

In any case, it needs a comment at the beginning of this file to give the context, and explain it's not to be used/updated/...

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+    return err
+
+
+class MockXPI(object):
+
+    def __init__(self, resources):
+        self.resources = resources
+
+    def info(self, name):
+        # Um...
```
magopian: ?

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+        return {"extension": name.split(".")[-1].lower()}
+
+    def read(self, name):
+        if isinstance(self.resources[name], bool):
+            return ""
```
magopian: Why an empty string?

magopian: (hint: comment ;)

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    Temporarily marks the library, framework, or version at the given path
+    as accepted. Used mainly for tests on now-banned, outdated Jetpack
+    versions.
+    """
+    dict_ = constants.LIBRARY_METADATA
+    for elem in path:
+        dict_ = dict_[elem]
+
+    unset = object()
```
magopian: Why use this? It doesn't seem idiomatic, and if it's a boolean, why not `dict_.get("banned", False)`?

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+    prev = dict_.get("banned", unset)
+    dict_["banned"] = False
+
+    yield
+
+    if prev is unset:
```
magopian: Oh, ok I understand now (cf precedent comment). I'd use `None` in that case.

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def assert_framework_files(err, files):
+    """Tests that all files are marked as pretested."""
+
+    identified_files = err.metadata.get("identified_files")
+    framework_files = err.metadata.get("framework_files")
+    assert identified_files and framework_files
+
+    for f in files:
```
magopian: No one letter variables please

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+
+        framework = err.metadata["framework"]
+
+        eq_(framework["id"], framework_id)
+
+        if data.get("banned"):
+            assert (("testcases_frameworks", "banned_framework") in
+                    (e["id"] for e in err.errors))
+        else:
+            # We don't bother reporting framework files for banned
```
magopian: That's a comment for the `if` case right?

tests/test_frameworks.py
```diff
@@ -0,0 +1,180 @@
...
+            # frameworks, since they're blanket rejected and won't
+            # see further review.
+
+            eq_(set(data["framework_files"]),
+                set(err.metadata["framework_files"]))
+
+            if data.get("unversioned"):
```
magopian: You're hardcoding the version below, so I'd also hardcode this branch (for the `besttoolbars` which is the only unversioned one in your fixture).

tests/test_jetpack.py
```diff
@@ -5,53 +5,60 @@
...
-        self.resources = resources
+    err = ErrorBundle()
+    content.test_packed_packages(err, xpi_package)
 
-    def read(self, name):
-        if isinstance(self.resources[name], bool):
-            return ""
-        return self.resources[name]
+    if allow_old_sdk:
+        ignore_ids = [
```
magopian: should it be named `ignore_id` (singular)?

tests/test_jetpack.py
```diff
@@ -5,53 +5,60 @@
...
 
-    def __iter__(self):
-        for name in self.resources.keys():
-            yield name
+        err.warnings = [e for e in err.warnings
+                        if e["id"] not in ignore_ids]
 
-    def __contains__(self, name):
-        return name in self.resources
+    print err.print_summary(verbose=True)
```
magopian: print?

tests/test_jetpack.py
```diff
@@ -5,53 +5,60 @@
...
 
 
 def test_not_jetpack():
     """Test that add-ons which do not match the Jetpack pattern are ignored."""
 
     err = _do_test(MockXPI({"foo": True, "bar": True}))
-    assert not err.errors
-    assert not err.warnings
-    assert not err.notices
+    eq_(err.errors, [])
```
magopian: why replace `assert` with `eq_` in this case? Maybe use `assert err.errors == []` if you want a precise comparison?

tests/test_jetpack.py
```diff
@@ -62,111 +69,94 @@ def test_bad_harnessoptions():
                             "components/harness.js": True,
                             "harness-options.json": "foo bar"}))
     assert err.failed()
-    assert err.warnings
+    assert err.errors
     print err.warnings
```
magopian: I would remove the print

tests/test_jetpack.py
```diff
@@ -62,111 +69,94 @@ def test_bad_harnessoptions():
...
-        bootstrap = bootstrap_file.read()
-
-    err = _do_test(MockXPI({"bootstrap.js": bootstrap,
-                            "harness-options.json":
-                                json.dumps(harnessoptions),
-                            "foo.png": True,
-                            "bar.JpG": True,
-                            "safe.GIF": True,
-                            "icon.ico": True,
-                            "foo/.DS_Store": True}))
```
magopian: You didn't keep the `foo/.DS_Store` file?

tests/test_jetpack.py
```diff
@@ -176,34 +166,32 @@ def test_pass_manifest_elements():
...
-    assert "identified_files" in err.metadata
-    assert "bootstrap.js" in err.metadata["jetpack_identified_files"]
-    assert "bootstrap.js" in err.metadata["identified_files"]
+    eq_(err.metadata["jetpack_loaded_modules"],
+        ["addon-kit-lib/drawing.js"])
+    assert_framework_files(err, ["bootstrap.js"])
 
-    assert "jetpack_unknown_files" in err.metadata
-    assert not err.metadata["jetpack_unknown_files"]
+    if False:
```
magopian: wat

tests/test_jetpack.py
```diff
@@ -328,20 +300,23 @@ def test_mismatched_db_hash():
...
+    assert "jetpack_loaded_modules" not in err.metadata
 
-    assert "jetpack_unknown_files" in err.metadata
-    unknown_files = err.metadata['jetpack_unknown_files']
-    nose.tools.eq_(len(unknown_files), 2)
-    nose.tools.ok_('bootstrap.js' in unknown_files)
-    nose.tools.ok_('resources/bootstrap.js' in unknown_files)
+    # There is currently no analog for this. Restore if we decide to
+    # create one.
+    # assert "jetpack_unknown_files" in err.metadata
```
magopian: Do we need this?

tests/test_jetpack.py
```diff
@@ -328,20 +300,23 @@ def test_mismatched_db_hash():
 
     err = _do_test(MockXPI({"bootstrap.js": bootstrap,
                             "resources/bootstrap.js": bootstrap,
-                            "harness-options.json":
-                                json.dumps(harnessoptions)}))
-    print err.print_summary(verbose=True)
-    assert not err.failed()
+                            "harness-options.json": harnessoptions}))
+    assert err.failed()
```
magopian: This is the exact opposite as what it was, is that on purpose?

tests/test_jetpack.py
```diff
@@ -456,23 +433,12 @@ def test_outdated_sdkversion():
...
+        err = _do_test(xpi, allow_old_sdk=False)
 
     assert err.failed()
     # Make sure we don't have any version mismatch warnings
+    eq_(err.errors, [])
     eq_(len(err.warnings), 1)
     eq_(err.warnings[0]["id"][2], "outdated_version")
 
-
-def test_future_sdkversion():
```
magopian: Don't we want this test anymore?

tests/test_libraries.py
```diff
@@ -0,0 +1,128 @@
...
+    validation process.
+    """
+
+    package_data = open("tests/resources/libraryblacklist/blocked.xpi")
+    package = XPIManager(package_data, mode="r", name="blocked.xpi")
+    err = ErrorBundle()
+
+    test_content.test_packed_packages(err, package)
+
+    print err.print_summary()
```
magopian: I don't think we want/need those `print`s

tests/test_libraries.py
```diff
@@ -0,0 +1,128 @@
...
+    Tests the validator's ability to hash each individual file and (based on
+    this information) determine whether the addon passes or fails the
+    validation process.
+    """
+
+    package_data = open("tests/resources/libraryblacklist/blocked.xpi")
+    package = XPIManager(package_data, mode="r", name="blocked.xpi")
+    err = ErrorBundle()
+
+    test_content.test_packed_packages(err, package)
```
magopian: I would create a helper function for the 4 lines above that seem to be used throughout this file. Also, I'd use the `with open(...) as ...` idiom.

tests/test_libraries.py
```diff
@@ -0,0 +1,128 @@
...
+        print err.print_summary()
+
+        assert not err.errors
+        assert err.notices
+        notice = err.notices[0]
+
+        eq_(notice["id"],
+            ("testcases_libraries", "detect_libraries",
+             "detected_library_message"))
+        assert all(m in notice["description"]
```
magopian: no one letter variables please

validator/errorbundler.py
```diff
@@ -384,20 +389,28 @@ def print_summary(self, verbose=False, no_color=False):
...
+                framework = self.metadata["framework"]
+
+                self.handler.write("\n")
+                self.handler.write("<<GREEN>>%s %s framework add-on detected."
+                                   "<<NORMAL>>\n" % (framework["name"],
+                                                     framework["version"]))
+
+            if self.metadata.get("framework_files"):
+                self.handler.write("Framework files:")
+                for filenamename in self.metadata["framework_files"]:
```
magopian: typo `filename`

validator/errorbundler.py
```diff
@@ -384,20 +389,28 @@ def print_summary(self, verbose=False, no_color=False):
...
+                    self.handler.write(" %s" % filenamename)
```
magopian: `self.handler.write(' '.join(self.metadata['framework_files']))`

validator/errorbundler.py
```diff
@@ -384,20 +389,28 @@ def print_summary(self, verbose=False, no_color=False):
...
+                self.handler.write("")
```
magopian: what does writing an empty string do?

validator/errorbundler.py
```diff
@@ -384,20 +389,28 @@ def print_summary(self, verbose=False, no_color=False):
...
+
+            if self.metadata.get("identified_files"):
+                self.handler.write("Identified files:")
+                for name, data in (self.metadata["identified_files"]
+                                    .iteritems()):
+                    self.handler.write(" %s:" % name)
+                    self.handler.write("\n".join("  %s %s (%s)" % tuple(s)
```
magopian: no single letter variable please. What's in `s`? Is the `tuple(s)` call needed?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+from validator.testcases import libraries
+from validator.version import Version
+
+
+def fake_version(files):
+    """
+    Returns a hash to act in place of a version number, given the contents
+    of a set of known files.
+    """
+    import hashlib
```
magopian: import at the top of the module

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+    return hashlib.sha256("".join(files)).hexdigest()[:16]
+
+
+class MetaFramework(type):
+    def __new__(mcls, name, bases, dict_):
+        cls = super(MetaFramework, mcls).__new__(mcls, name, bases, dict_)
+        if cls.NAME:
```
magopian: It seems this is only to not list the base `Framework` class, so maybe have a specific case just for it, `if cls.__name__ != 'Framework':`

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+        # Update certain metadata from the metadata JSON.
+        # Ideally this would also update based on data for the
+        # specific version of the framework, but there are
+        # bootstrapping issues.
+        for prop in "framework_files", "framework_prefixes":
+            if prop in self.framework_metadata:
+                setattr(self, prop.upper(),
+                        tuple(self.framework_metadata[prop]))
+
+    NAME = None
```
magopian: All properties together at the beginning of the class please

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+    FRAMEWORK_FILES = ()
+
+    # Tuple of file path prefixes for which matching files, if they
+    # exist, belong to the framework.
+    FRAMEWORK_PREFIXES = ()
+
+    @cached_property
+    def ID(self):
+        """Internal ID of the framework."""
+        return self.NAME and self.NAME.lower()
```
magopian: This will return `False` if there's no `self.NAME`, and not an empty string or `None`. I understand this is only missing for the base `Framework` class, so maybe you don't need this test?

validator/testcases/frameworks.py
magopian: Ah, I see you're using the `NAME` property anyway, so I guess it's good this way

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def detect(self):
+        """Returns true if the add-on matches the given framework."""
+        return False
```
magopian: `raise NotImplementedError`?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+
+        return LIBRARY_METADATA["frameworks"].get(self.ID, DEFAULT)
+
+    @cached_property
+    def framework_files(self):
+        """
+        Returns a list of files in the package which belong to the framework
+        in question. Files may or may not match known versions of the file.
+        """
+        return (filter(self.xpi.__contains__, list(self.FRAMEWORK_FILES)) +
```
magopian: why the call to `list`, it says from the comment https://github.com/mozilla/amo-validator/pull/271/files#diff-247aeb318317091e4843b00ac0f88b12R60 that it's a tuple?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+        applicable, and the contents are sane. Otherwise returns `None`.
+        """
+        if self.METADATA_FILE and self.METADATA_FILE in self.xpi:
+            try:
+                metadata = json.loads(self.xpi.read(self.METADATA_FILE))
+                if isinstance(metadata, dict):
+                    return metadata
+            except:
+                pass
+        return None
```
magopian: Not needed

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+        framework files.
+        """
+        return fake_version(self.xpi.read(f)
+                            for f in sorted(self.framework_files)
+                            if self.CHECKED_FILES.search(f))
+
+    @cached_property
+    def version(self):
+        """
+        Returns the version of the framework used, or `None` if not known.
```
magopian: I don't think it's returning `None`, but a fake version, right?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+            self.error(
+                err_id=("testcases_frameworks",
+                        "invalid_metadata"),
+                error="Invalid metadata file",
+                description=("This add-on does not contain a valid metadata "
+                             "file for the %s framework."
+                             % self.NAME),
+                filename=self.METADATA_FILE)
+
+            return self.valid
```
magopian: Shouldn't it set `self.valid = False` first?

magopian: Ah, it's done in `self.error`, my bad

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+
+        messages = (metadata.get("messages", [])
+                    + version_metadata.get("messages", []))
+
+        banned = metadata.get("banned")
+        if not self.ACCEPTED or banned:
+            report = self.err.warning if banned == "maybe" else self.error
+            report(
+                err_id=("testcases_frameworks",
+                        "banned_framework"),
```
magopian: move on the line above

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+                message="Banned add-on framework in use",
+                description=(["This extension makes use of the %s framework, "
+                              "which is not currently accepted for use on AMO"
+                              % self.NAME]
+                             + messages))
+
+        elif not version_metadata:
+            self.error(
+                err_id=("testcases_frameworks",
+                        "unknown_framework_version"),
```
magopian: Does it fit on the previous line?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+            identity = identified.get(path)
+            if not identity:
+                # This may turn out to be something we want:
+                # Skip checks on framework files even if not identified
+                # by hash.
+                # identified[path] = {"banned": True}
+
+                self.error(
+                    err_id=("testcases_frameworks",
+                            "unknown_file"),
```
magopian: on the previous line

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+                    error="Unknown framework file",
+                    description=("A file belonging to the %s framework does "
+                                 "not match any known checksum." % self.NAME),
+                    filename=path)
+
+            elif not any(match[0:2] == tag for match in identity["sources"]):
```
magopian: comment

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+        if not super(Jetpack, self).validate():
+            return self.valid
+
+        # Legacy
+        self.err.metadata["is_jetpack"] = True
+
+        return self.valid
+
+
+# Built on top of Jetpack, so must come first.
```
magopian: How do you enforce that? I thought you would have needed to put it before Jetpack in the file?

magopian: Ah, sorry, I understand, the Jetpack class above is an "abstract" one (with `NAME = None`). I would add a docstring.

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def detect(self):
+        return any(f in self.xpi
+                   for f in (
+                       "chrome/content/CrossriderEXT.js",
+                       "chrome/content/crossrider.js",
+                       "chrome/content/crossriderapi.js"))
+
+
+class Jetpack(Framework):
```
magopian: Add a docstring explaining that it's an "abstract" (or "base") framework, with `NAME = None`

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+                    filename=zip_path)
+
+            # We aren't going to keep track of anything that isn't an official
+            # Jetpack file.
+            identity = identified_files.get(zip_path)
+            if identity and "jetpack" in (s[0] for s in identity["sources"]):
+                # Keep track of all of the valid modules that were loaded.
+                loaded_modules.append("".join([module["packageName"], "-",
+                                               module["sectionName"], "/",
+                                               module["moduleName"], ".js"]))
```
magopian: This is a strange way to do string interpolation, why not `%s-%s/%s.js`?

validator/testcases/frameworks.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def detect(self):
+        return isinstance(self.metadata, dict)
+
+    FRAMEWORK_FILES = ("bootstrap.js",
+                       "scriptify.json")
+
+    @cached_property
+    def version(self):
+        return self.metadata.get("scriptify-version")
```
magopian: you're using `.get()` here, and `[...]` in the previous framework.

validator/testcases/installrdf.py
```diff
@@ -30,16 +28,16 @@ def _test_rdf(err, install):
     """Wrapper for install.rdf testing to make unit testing so much
     easier."""
 
-    shouldnt_exist = ("hidden", )
+    shouldnt_exist = "hidden",
```
magopian: To be honest, I'd rather `["hidden"]` for perfect readability

validator/testcases/installrdf.py
```diff
@@ -50,19 +48,18 @@ def _test_rdf(err, install):
                       "multiprocessCompatible",
                       "optionsType",
                       "type",
-                      "updateInfoURL",
```
magopian: deprecated?

validator/testcases/libraries.py
```diff
@@ -0,0 +1,204 @@
...
+        for lib, data in metadata[lib_type].iteritems():
+            for version, vdata in data.get("versions", {}).iteritems():
+                banned = False
+                if lib_type == "libraries":
+                    # Currently, banned library versions propagate to
+                    # files, banned framework versions do not. We may need
+                    # greater granularity in the future.
+                    banned = merge_banned(data, vdata)
+
+                for path, hash in vdata.get("files", {}).iteritems():
```
magopian: `hash` is a python builtin, so rather use `hash_` for example

validator/testcases/libraries.py
```diff
@@ -0,0 +1,204 @@
...
+                res[lib] = res[lib].intersection(versions)
+
+    def max_version(versions):
+        return str(max(map(Version, versions)))
+
+    if library:
+        if res and library in res:
+            return max_version(res[library])
+    else:
+        match = next(((k, v) for k, v in res.iteritems() if v),
```
magopian: This deserves a comment

validator/testcases/libraries.py
```diff
@@ -0,0 +1,204 @@
...
+                     None)
+        if match:
+            return match[0], max_version(match[1])
```
magopian: I would have written:
```python
matches = [(k, v) for k, v in res.iteritems() if v]
if matches:
    match = matches[0]
    return match[0], max_version(match[1])
```

Simpler, idiomatic, explicit, more readable.

magopian:
Thank you for your work, it is impressive.

In the future, please consider:
1/ multiple atomic commits
2/ multiple PRs if it's fixing unrelated things (eg: rewriting or improving some code to clean it is ok, completely rewriting huge blobs isn't)
3/ committing "file moves" and "code modifications" separately, to be able to only review the latter
4/ add comments and docstring as much as possible, this really helps understanding the code, reviewing it, and maintaining it in the future
5/ document the rationale, the changes... in the PR description. The discussion we had in person was very helpful for that (understanding the flow of lib updates, storing the hashes in olympia's database...)
6/ code simplicity. This is very important for maintainability. Code must be as clear, readable and straightforward as possible. Avoid complex and convoluted solutions at all costs, even if they are more concise or "more clever". In fact, avoid it in particular if they are.

Regarding the PR itself, as discussed, it would make sense to distribute the lib fetcher separately in the future, maybe even release it in a generic enough way so it may be useful to people outside of Mozilla?
Also, I don't think the code to update olympia's codebase with the hashes from the latest versions of the libs is in this PR, so if possible, i'd like:
1/ a separate bug/PR for the new lib updater code (not necessarily as a separate repo yet)
2/ a separate bug/PR to update olympia's database (using the web API if I remember correctly what you wanted to do from our discussion at Whistler?)
3/ maybe do not implement 2/ for now, as that would:
    - avoid having to wait for the fix of the oauth code on olympia
    - still let us run a cron job to take a local file and populate it in the DB (that would even be easier to implement)

validator/validate.py
```diff
@@ -31,7 +31,9 @@ def validate(path, format="json",
         The format to return the results in. Defaults to "json". Currently, any
         other format will simply return the error bundle.
     `approved_applications`:
-        Path to the list of approved application versions
+        Path to the list of approved application versions.
```
magopian: Document that it can be a dict (same for the `library_metadata` below).

Should we remove the `app_versions.json` file from the repository now that it's not used anymore? That would be convenient, as we won't have to maintain it manually anymore

tests/resources/bootstrap-jpm.js
kmaglione: This is an unmodified file from the add-on SDK. It needs to be included as-is so the stock checksums will match it.

tests/test_frameworks.py
kmaglione: Sorry, the comment was because I don't understand why this method even exists. I'll remove it.

tests/test_frameworks.py
kmaglione: I moved this from another file. Essentially, whoever implemented it decided that `True` should be used for files which exist but may be empty, rather than just `''`.

tests/test_frameworks.py
kmaglione: It's not a boolean. If the property doesn't exist, it needs to be removed entirely after the test, so `unset` is a sentinel. I thought I'd added a comment to that effect.

tests/test_frameworks.py
kmaglione: Yeah. Donno how it wound up in the `else` branch.

tests/test_frameworks.py
kmaglione: Hm. The version is generated from the contents of the files. I don't really want to hard code it, since that would rely on the behavior of `fake_version`, or the contents of the stub files, not changing, which I don't want to depend on. I just want to make sure that it comes out to the same as the return value for `fake_version`.

kmaglione: I replaced the hard coded `"42"` with a `VERSION` constant.

tests/test_jetpack.py
kmaglione: If I changed it to a tuple rather than a list, and changed the `in` to `==`. I implemented it this way since it didn't seem to add significant complexity, and it seemed likely we'd want to add other IDs in the future.

tests/test_jetpack.py
kmaglione: Because nose doesn't have any special behavior for reporting assert failures, and if this fails, I'd like to see what value it wound up with, without having to add debugging code. I changed this because it was failing, and I needed that information.

tests/test_jetpack.py
kmaglione: No. That file generates a warning, and there didn't seem to be a good reason to keep it.

tests/test_jetpack.py
kmaglione: See comment. I can comment out the old lines of code, rather than `if False`ing them, if you prefer.

tests/test_jetpack.py
kmaglione: Yes. There was no warning for this in the old code. The new code reports files which are expected to match a known hash but don't.

tests/test_jetpack.py
kmaglione: I don't think so. I was just leaving it there for reference in case it turns out that someone was using that property.

tests/test_jetpack.py
kmaglione: No. The expected behavior is that we should warn when a future version of the SDK is used. But this test wasn't actually testing that, since 1.8 had not been a future version in a very long time.

tests/test_libraries.py
kmaglione: This has been used by the majority of validator tests for a long time. The output doesn't show unless the test fails, and when the test fails, it usually turns out to be useful.

tests/test_libraries.py
kmaglione: I agree. I did that with `test_jetpack.py`, I just didn't want to refactor all of these. This file was renamed from `test_libraryblacklist.py`, so most of this code is not new.

validator/errorbundler.py
kmaglione: This function writes newlines after the string. This just writes out a newline.

validator/errorbundler.py
kmaglione: It'd need to be something like `"\n ".join(...)`, but I'm not sure of all of the implications of writing output that way with this code. I was mainly just updating the current code to use the new data structures.

validator/errorbundler.py
kmaglione: `s` is a list describing the source. Without the `tuple()`, it's processed as a single argument.

validator/testcases/frameworks.py
kmaglione: Because we can't concatenate a tuple to a list, and `filter` will return a tuple if it gets a tuple for input.

validator/testcases/frameworks.py
kmaglione: Yeah. Behavior changed.

validator/testcases/frameworks.py
kmaglione: Sorry. It's  a JS thing.

kmaglione: Oh, also this is old code from `jetpack.py`

validator/testcases/frameworks.py
kmaglione: Because it's required to be present in Kango, but not in Scriptify.

validator/testcases/installrdf.py
kmaglione: No, it never actually existed in install.rdf.

validator/validate.py
kmaglione: It's still used for now, when the validator is run standalone. Ideally, I'd like to be able to fetch it on demand from AMO, in the future, but at the moment we still need it.

validator/testcases/installrdf.py
kmaglione: I'm going to remove these changes for now. I'm not sure how they wound up in this PR.

kmaglione:
> 2/ multiple PRs if it's fixing unrelated things (eg: rewriting or improving some code to clean it is ok, completely rewriting huge blobs isn't)

I think the only unrelated things that snuck in here were in `installrdf.py`. I'm going to pull those changes out and put them into a separate PR.

> 3/ committing "file moves" and "code modifications" separately, to be able to only review the latter

I think there was only one file that actually moved here (`test_libraryblacklist.py`). The code from `jetpack.py` that I didn't delete actually got merged into `frameworks.py` in a way that I don't think rename tracking would have helped. I'll try to see if I can fix the former with a rebase.

> 6/ code simplicity. This is very important for maintainability. Code must be as clear, readable and straightforward as possible. Avoid complex and convoluted solutions at all costs, even if they are more concise or "more clever". In fact, avoid it in particular if they are.

I think you're going to point these out when you see them. I try to keep most of my code as simple as possible, so we probably have different definitions.

> 1/ a separate bug/PR for the new lib updater code (not necessarily as a separate repo yet)

I think I'll move them to a separate repo.

> 2/ a separate bug/PR to update olympia's database (using the web API if I remember correctly what you wanted to do from our discussion at Whistler?)

I have a separate repo for this, based on `andymckay/amo-oauth`. I need to do some cleanup, then I'll put it up somewhere for review.

> 3/ maybe do not implement 2/ for now, as that would:
> - avoid having to wait for the fix of the oauth code on olympia
> - still let us run a cron job to take a local file and populate it in the DB (that would even be easier to implement)

I'm not sure a cronjob be sufficient here, since we need to add manual hashes fairly often. It might be sufficient for the short term to split out the oauth changes and add a management command so someone from ops can update the hashes in the short term, though. I'd really rather not wait too long.

kmaglione:
I think this is everything.

I also removed the metadata updater and some unrelated changes that originally got sucked in, moved the metadata validation code that was into Olympia into `libraries.py` and `test_libraries.py`, and factored out some common code from a few test files.

kmaglione:
It turns out I forgot a couple of things, actually. I have an update which changes `"banned": (bool|string)` to `"deprecated": int`, and updates the README. It's marooned on my office computer at the moment, since it seems to have gotten kicked off the network. Will update later today.

kmaglione:
Looks like I forgot to push this last week. This is now mergeable again, and I think it addresses everything.

tests/test_jetpack.py
magopian: You're right, the validator isn't using pytest (yet? :)

tests/test_jetpack.py
magopian: Yes please ;) I'd rather have commented out code (even though I dislike it a lot, if it's commented, it's not used, if it's not used, it shouldn't be there) than this construct.

validator/testcases/frameworks.py
magopian: understood, thanks

magopian:
A few comments aren't answered yet (eg https://github.com/mozilla/amo-validator/pull/271#discussion_r35101643 and https://github.com/mozilla/amo-validator/pull/271#discussion_r33466960).

Regarding your [last comment](https://github.com/mozilla/amo-validator/pull/271#issuecomment-119038083):
1/ you said "I think I'll move them to a separate repo.": is that done? If so, shouldn't the documentation (in the README, about the new data structure for libs/frameworks hashes) be in this new repo instead?
2/ your answer about the cronjob not being sufficient: I have a bad feeling about this, it really looks over-engineered.

I understand your need to have a simpler (not "blocked" by the filesystem/repo) way to update the hashes. In that case, the "usual" solution would be to create an admin model for that. Or a management command, yes.

This PR is really large, I can't remember what's in there, and what isn't, can you please sum it up? Is it "only" the code to use the new "lib/framework hashes data structure"? Does it still have the code related to the lib updating? Or the "posting to AMO to update its data"?

kmaglione:
> 1/ you said "I think I'll move them to a separate repo.": is that done?

Yes, but I haven't made it public yet.

> If so, shouldn't the documentation (in the README, about the new data structure for libs/frameworks hashes) be in this new repo instead?

I think it makes sense to have it in this repo, since the validator deals with it extensively.

> 2/ your answer about the cronjob not being sufficient: I have a bad feeling about this, it really looks over-engineered.
>
> I understand your need to have a simpler (not "blocked" by the filesystem/repo) way to update the hashes. In that case, the "usual" solution would be to create an admin model for that. Or a management command, yes.

That would be the usual solution, yes. The problem with that solution, though, is that all of the admin pages we currently have are kludgy and poorly-maintained. This one would need to deal with managing much more complex data, would probably have to be updated more often, would have to be updated in sync with the validator, would most likely have to include an incredible amount of logic for mapping DB models to data structures the validator deals with, would require a local tool anyway for hashing libraries and frameworks that need to be manually approved, would then require either uploading those new hashes to a running server and re-downloading the libraries database to test, or otherwise re-implementing the same logic again in a local client...

And, as for a management command, that would mean needlessly relying on someone from ops every time we need to add a hash, or change a flag.

In either case, we'd still need to implement all of the same logic. And on the server side, I think it would require much more complexity, not less.

At any rate, I'm not ruling out moving more of the logic to Olympia, but I'd prefer to go live with the current solution first, and talk about what needs to change if it doesn't appear to be working well.

> This PR is really large, I can't remember what's in there, and what isn't, can you please sum it up?

The current incarnation has:

* Code for detecting the following frameworks, and what files belong to the framework:

  - BestToolbars
  - Conduit
  - Crossrider
  - Jetpack (CFX)
  - Jetpack (JPM)
  - OpenForge
  - Kango
  - Scriptify

* New code for detecting known libraries with the added ability to:

  - Mark libraries, frameworks, or versions as banned.
  - Add warnings for particular libraries ("Prototype.js may not be used in browser overlays.").
  - Warn when a file which should belong to a framework is unknown.

The current version of the patch I have for AMO leaves out the API code for the moment, and just has an admin page with a file upload field. I think that should be sufficient in the short term.

In the medium term, we're going to need the API fixes in my original PR anyway if we're actually going to deliver on our promise of a signing API.

magopian:
Ok, here's the deal. This PR has grown so huge that I can't (and don't want to) go over it again. It includes so many different changes, some unrelated, that I just can't grok it.

On the other hand, I don't want to reject it "just because I'm too lazy to go over it (again)", and I know now (after some discussion with you on irc) that at least some of this work is needed for the future of the validator (and dealing with the chrome-style add-ons API). Also, you're the only current active maintainer, and you have agreed on a plan with management, if I understand things correctly.

So unless you find someone who's ready to review it, I'll just say "merge it while I'm looking the other way".

To be honest, I'm pretty annoyed at this whole thing, and would like this to not happen again in the future. If there's anything I can do to help this not happening again, please let me know.

muffinresearch:
I'm thinking if it would be useful to focus on capturing these changes on the addons-validator project - we're going to need to cover the same processes, but as there's clearly improvements to the amo-validator here it would be nice to get them rather than us just port what's there. 

A starting point might be to add some issues breaking down what's in this patch on the addons-validator project so we can address them as we bring across the JS lib/framework detection?

EnTeQuAk:

PR #272, opened 2015-02-17
Gecko 35 tests are really for Gecko 36

mstriemer:


magopian:
r+, thanks @mstriemer !

PR #273, opened 2015-03-03
Add the langpack reference XPIs to the package (bug 1138662)

magopian:
Fixes [bug 1138662](https://bugzilla.mozilla.org/show_bug.cgi?id=1138662)


magopian:
@kmaglione r?

Do you know, by any chance, if there's anything else missing? Is there a way to check?

kmaglione:
I just diffed the file list from PyPI with the repo, and it doesn't look like anything else important is missing.

I don't really have a strong opinion on adding the entire testcases folder. It probably shouldn't be necessary. But it would be nice if we could somehow run tests against the generated package before publishing so we catch these kinds of packaging errors automatically.

Anyway, this looks good for now. r+

PR #274, opened 2015-03-11
Gecko 37 compat

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1134367
https://bugzilla.mozilla.org/show_bug.cgi?id=1134362

### DIFF
--- a/validator/testcases/javascript/instanceactions.py
+++ b/validator/testcases/javascript/instanceactions.py
@@ -13,7 +13,8 @@ node
 import actions
 from validator.compat import (FX10_DEFINITION, FX29_DEFINITION,
                               FX30_DEFINITION, FX31_DEFINITION,
-                              FX33_DEFINITION, FX36_DEFINITION)
+                              FX33_DEFINITION, FX36_DEFINITION,
+                              FX37_DEFINITION)
 from validator.constants import BUGZILLA_BUG, MDN_DOC
 from instanceproperties import _set_HTML_property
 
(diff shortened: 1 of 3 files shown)


validator/testcases/javascript/instanceactions.py
```diff
@@ -444,6 +457,7 @@ def os_file_readTo_callback(argument, traverser, node, wrapper):
     "addLivemark": livemarkCallback,
     "removeLivemark": livemarkCallback,
     "getLivemark": livemarkCallback,
+    "quote": quote_callback,
```
mstriemer: This tells the traverser to call `quote_callback` anytime a `quote` method is called. This means that it is unfortunately not just for String arguments but any object. Since we aren't actually running the code we don't really have a good way of knowing if this is a String or not.

validator/testcases/javascript/instanceactions.py
```diff
@@ -422,6 +423,18 @@ def os_file_readTo_callback(argument, traverser, node, wrapper):
         tier=5)
 
 
+def quote_callback(argument, traverser, node, wrapper):
+    traverser.warning(
+        err_id=("js", "string", "quote"),
+        warning="The quote interface has been removed",
+        description="The quote interface has been removed. See %s for more "
```
mstriemer: This should probably mention that the interface was removed from `String.prototype`.

validator/testcases/regex.py
```diff
@@ -1546,6 +1547,23 @@ def tests(self):
...
 
+@register_generator
+class Gecko37RegexTests(CompatRegexTestHelper):
+    """Regex tests for Gecko 37 updates."""
+
+    VERSION = FX37_DEFINITION
+
+    def tests(self):
+        yield self.get_test(
+            r"\bnsIDownloadManagerUI\b",
```
mstriemer: This is just a regex test so if this appears anywhere in the code then the warning/error will be registered. This unfortunately includes comments. It might be nice to run the code through something that strips the comments out first.

mstriemer:
I'm not sure what's up with the functional tests. They passed locally for me.

PR #275, opened 2015-03-12
Bug 1142732 - Support computed property names in object literals.

kmaglione:


validator/testcases/regex.py
```diff
@@ -233,14 +233,12 @@ def applicable(cls, err, filename, document):
     def js_tests(self):
         title = "Marionette access is disallowed"
         for ref, bug in self.MARIONETTE_REFERENCES.items():
-            print ref
```
magopian: :+1:

magopian:
Have you tried this on the code? Could you please add a test to the PR demonstrating this? Thanks!

mstriemer:
Is there a test you can add for this? I'm hitting the error this fixes in https://github.com/mozilla/amo-validator/pull/274 but that seems like more of a side-effect than a real test.

kmaglione:
I have tested it, but I'll add an actual test.

kmaglione:
Added test.

jetpack/addon-sdk
```diff
@@ -1 +1 @@
-Subproject commit 74d7ae8f5496979baf1b7e995b38474dc81d5e70
+Subproject commit e2b154aff5c452d3b4f25bd67513122bd6172319
```
mstriemer: Does this need to change? Looks unrelated.

mstriemer:
Thanks for adding the test. I'm r+ other than the addon-sdk change, is that needed?

kmaglione:
Ugh. I didn't mean to add that. I made the changes on the wrong branch. I guess `stash` added to the index.

kmaglione:
Fixed SDK submodule change.

mstriemer:
Looks like the change to the `jsshell` version is missing, was that in here before? Something like https://github.com/mozilla/amo-validator/pull/274/files#diff-354f30a63fb0907d4ad57269548329e3R10.

kmaglione:
Ugh. wth...

kmaglione:
OK, I think it's all there now.

PR #277, opened 2015-03-14
Bug 666066: Show warnings for obfuscated/minified code.

kmaglione:


kmaglione:
Looks like this needs #275 to build.

magopian:

kmaglione:
I rebased on top of #278, since it seemed likely there would be conflicts.

validator/testcases/javascript/traverser.py
```diff
@@ -15,6 +15,7 @@
 IGNORE_POLLUTION = False
 POLLUTION_COMPONENTS_PATH = re.compile(r"/?components/.*\.jsm?")
 POLLUTION_EXCEPTIONS = set(["Cc", "Ci", "Cu", ])
+MAX_LINE_COMPLEXITY = 100
```
magopian: That looks like a lot. I was expecting something like 10. Are you afraid of false positives?

validator/testcases/javascript/traverser.py
```diff
@@ -83,6 +85,22 @@ def run(self, data):
             raise
         self._debug("END>>")
 
+        for line in sorted(self.complexity.keys()):
```
magopian: Why sorting?

magopian:
There's nothing in there to detect code obfuscation, right? Also, curious why you didn't go with the "count number of `;` per line" method that seemed easier.

It's missing tests.

validator/testcases/javascript/traverser.py
magopian: Following in-person discussion, seems that 100 is just right.

validator/testcases/javascript/traverser.py
magopian: After in-person discussion, it's so the warnings are displayed in the correct order (the key is the line number)

magopian:
As discussed in-person: needs tests, and then it's good to merge

r+wc

validator/testcases/javascript/traverser.py
muffinresearch: @magopian Can you share the rationale here too so it's documented? We might want similar checks in the JS validator.

magopian: Oh, wow, I should have noted that at the time, I completely forgot :'( @kmaglione could you please explain it again?

EnTeQuAk:

PR #278, opened 2015-03-14
Bug 1143247: Add support for validating the contents of arrow function e...

kmaglione:
...xpressions.


diox:
This needs a test

kmaglione:
Added tests.

magopian:

tests/test_js_instanceactions.py
```diff
@@ -101,19 +101,26 @@ def test_callexpression_argument_traversal():
...
-
-    assert _do_test_raw("""
-    function foo(x){}
-    foo({"bar":function(){
-        eval("evil");
-    }});
-    """).failed()
+    DECLARATIONS = (
+        "function foo(x){}",
+        "var foo = function foo(x){}",
```
magopian: Seems this one was missing previously, good catch :+1:

validator/testcases/javascript/nodedefinitions.py
```diff
@@ -57,6 +57,9 @@ def node(branches=None, dynamic=False, action=None, returns=False,
     "FunctionExpression": node(branches=("body", ), dynamic=True,
                                action=actions._func_expr, returns=True,
                                is_block=True),
+    "ArrowExpression": node(branches=("body", ), dynamic=True,
+                            action=actions._func_expr, returns=True,
+                            is_block=True),
```
magopian: It's exactly the same as the `FunctionExpression`, what's its use?

tests/test_js_syntax.py
```diff
@@ -56,3 +56,42 @@ def test_spidermonkey_warning():
...
+        "if (true) ; else { %s }",
+        "while (true) { %s }",
+        "do { %s } while (true)",
+        "for (;;) { %s }",
+        "try { %s } catch (e) {}",
+        "try {} catch (e) { %s }",
+        "try {} finally { %s }",
+        "try {} catch (e if %s) {}",
+        # This is broken.
+        # "let (foo = bar) { %s }",
```
magopian: Is that a `TODO`? Do you want to create a new bug if it needs to get fixed, and add that bug number to the comment?

tests/test_js_syntax.py
```diff
@@ -56,3 +56,42 @@ def test_spidermonkey_warning():
...
+    Tests that blocks of code are actually evaluated under normal
+    circumstances.
+    """
+
+    ID = ("js", "traverser", "dangerous_global")
+
+    EVIL = "eval(evilStuff)"
+    BLOCKS = (
+        "function foo() { %s; }",
+        "function foo() { %s; yield 1; }",
```
magopian: What's the use of this `yield 1;` here (and below)?

magopian: Ah, I see from the MDN docs that it's legacy generator functions, TIL ;)

magopian:
r+wc

tests/test_js_syntax.py
kmaglione: It's https://bugzil.la/1009181, but support for `let` blocks has finally been removed, so I'll just remove the comment.

validator/testcases/javascript/nodedefinitions.py
kmaglione: For the purposes of the validator, it's exactly the same. There are some semantic differences, the main ones being:

* It captures the `this` object of its enclosing scope.
* It does not have access to the `arguments` object.

The main use, though, is that it's just shorter to type.

kmaglione:
Rebased and removed outdated comment.

PR #280, opened 2015-04-17
[WIP] Add support for validating unlisted add-ons as gatekeeper for auto...

kmaglione:
...mated signing.

Submitting early WIP with some of the required tests for feedback/testing Olympia code.

(CC @magopian)

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -1,3 +1,4 @@
+import collections
 import sys
 
 from validator.submain import populate_chrome_manifest
@@ -78,30 +79,45 @@ class TestCase(object):
             self.err.detected_Type = self.detected_type
 
     def assert_failed(self, with_errors=False, with_warnings=None):
-        """First, asserts that the error bundle registers a failure
-        (recognizing whether warnings are acknowledged). Second, if
-        `with_errors`is True, the presence of errors is asserted. If it is not
-        true (default), it is tested that errors are not present. If
-        `with_warnings` is not None, the presence of warnings is tested just
-        like `with_errors`.
+        """
+        Asserts that the error bundle has registered a failure. If
+        `with_warnings` is any true value, or `None`, a warning is
+        considered a failure.
 
+        `with_warnings` or `with_errors` may be any of the following:
+
+        * True: Messages of this type must be present.
+        * False: Messages of this type must not be present.
+        * None: Messages of this type may or may not be present.
+        * Iterable of dicts: For dict returned by the iterator, at least
+        one message must have a matching item for every key/value pair in the
+        dict.
         """
         assert self.err.failed(
             fail_on_warnings=with_warnings or with_warnings is None), \
-                "Test did not fail; failure was expected."
+            "Test did not fail; failure was expected."
 
-        if with_errors:
-            assert self.err.errors, "Errors were expected."
-        elif self.err.errors:
-            raise AssertionError("Tests found unexpected errors: %s" %
-                                self.err.print_summary(verbose=True))
+        def find_message(messages, props):
+            # Returns true if any message in messages has all of the
+            # key/value pairs in props.
+            return any(set(props.iteritems()) <= set(message.iteritems())
+                       for message in messages)
 
-        if with_warnings is not None:
-            if with_warnings:
-                assert self.err.warnings, "Warnings were expected."
-            elif self.err.warnings:
-                raise ("Tests found unexpected warnings: %s" %
-                           self.err.print_summary())
+        def test_messages(mtype, expected):
+            messages = getattr(self.err, mtype)
+
+            if isinstance(expected, collections.Iterable):
+                assert all(find_message(messages, props)
+                           for props in expected)
+            elif expected:
+                assert messages, "Expected %s." % mtype
+            elif expected is not None:
+                assert not messages, ("Tests found unexpected %s: %s" %
+                                      mtype,
+                                      self.err.print_summary(verbose=True))
+
+        test_messages('errors', with_errors)
+        test_messages('warnings', with_warnings)
 
     def assert_notices(self):
         """Assert that notices have been generated during the validation
--- a/tests/resources/installrdf/shouldnt_exist.rdf
+++ b/tests/resources/installrdf/shouldnt_exist.rdf
@@ -8,6 +8,7 @@
 	<em:type>2</em:type>
 	
 	<em:updateURL>true</em:updateURL>
+	<em:updateKey>meh</em:updateKey>
 	
 	<em:id>bastatestapp1@basta.mozilla.com</em:id>
 	<em:version>1.2.3.4</em:version>
--- a/tests/test_js_functions.py
+++ b/tests/test_js_functions.py
@@ -57,6 +57,38 @@ def test_createElement():
     document.createElement();
     """)
 
+def test_enablePrivilege():
+    """Tests that enablePrivilege throws a warning."""
+
+    err = _do_test_raw("""
+        netscape.security.PrivilegeManager.enablePrivilege()
+    """)
+    assert err.warnings
+
+    eq_(err.warnings[0]["id"],
+        ("js", "traverser", "dangerous_global"))
+    eq_(err.warnings[0]["signing_severity"], "high")
+
+def test_privileged_unprivileged_interaction():
+    """Tests that functions which may lead to privilege escalation are
+    detected."""
+
+    for meth in "cloneInto", "exportFunction":
+        err = _do_test_raw("""
+            Cu.%s(foo)
+        """ % meth)
+        assert err.warnings
+
+        eq_(err.warnings[0]["id"],
+            ("js", "traverser", "dangerous_global"))
+        eq_(err.warnings[0]["signing_severity"], "low")
+
+    for form in ("var obj = { __exposedProps__: {} };",
+                 "var obj = {}; obj.__exposedProps__ = {};"):
+        err = _do_test_raw(form)
+        assert err.warnings
+        eq_(err.warnings[0]["signing_severity"], "high")
+
 def test_synchronous_xhr():
     "Tests that syncrhonous AJAX requests are marked as dangerous"
 
--- a/tests/test_js_signing.py
+++ b/tests/test_js_signing.py
@@ -1,0 +1,361 @@
+"""
+Various tests related to validation for automated signing.
+"""
+
+from .helper import RegexTestCase
+from .js_helper import TestCase
+
+
+class TestSearchService(TestCase, RegexTestCase):
+    """Tests that warnings related to the search service trigger warnings."""
+
+    def test_changes(self):
+        """Tests that setting changes trigger warnings."""
+
+        def test(obj, prop, stuff, warnings):
+            self.setUp()
+            self.run_script("""%s.%s%s;""" % (obj, prop, stuff))
+            self.assert_failed(with_warnings=warnings)
+
+        objs = ("Cc[''].getService(Ci.nsIBrowserSearchService)",
+                "Services.search")
+
+        for obj in objs:
+            warnings = {"signing_severity": "high",
+                        "id": ("testcases_javascript_actions",
+                               "search_service",
+                               "changes")},
+
+            for prop in "currentEngine", "defaultEngine":
+                yield test, obj, prop, " = foo", warnings
+
+            warnings[0]["signing_severity"] = "medium"
+
+            for meth in ("addEngine", "addEngineWithDetails",
+                         "removeEngine", "moveEngine"):
+                yield test, obj, meth, "(foo, bar, baz)", warnings
+
+
+    def test_registry_write(self):
+        """Tests that Windows registry writes trigger warnings."""
+
+        warnings = ({"id": ("testcases_javascript_actions",
+                            "windows_registry", "write"),
+                     "signing_severity": "medium"},
+                    {"id": ("js", "traverser", "dangerous_global"),
+                     "signing_severity": "low"})
+
+        def test(method):
+            self.setUp()
+            self.run_script("""
+                Cc[""].createInstance(Ci.nsIWindowsRegKey).%s(foo, bar);
+            """ % method)
+            self.assert_failed(with_warnings=warnings)
+
+        for method in ("create", "createChild", "writeBinaryValue",
+                       "writeInt64Value", "writeIntValue", "writeStringValue"):
+            yield test, method
+
+    def test_evalInSandbox(self):
+        """Tests that evalInSandbox causes signing warnings."""
+
+        self.run_script("""
+            Cu.evalInSandbox("foobar()", sandbox);
+        """)
+        self.assert_failed(with_warnings=[{"signing_severity": "low"}])
+
+
+    def test_pref_branches(self):
+        """
+        Tests that writes to potentially dangerous preferences are flagged.
+        """
+
+        def test(pref, severity):
+            warnings = [
+                {"message": "Potentially unsafe preference branch referenced",
+                 "signing_severity": severity}]
+
+            self.setUp()
+            self.run_script("""
+                Services.prefs.setCharPref('%s', '42');
+            """ % pref)
+            self.assert_failed(with_warnings=warnings)
+
+        PREFS = (("browser.newtab.url", "high"),
+                 ("browser.newtabpage.enabled", "high"),
+                 ("browser.search.defaultenginename", "high"),
+                 ("browser.startup.homepage", "high"),
+                 ("keyword.URL", "high"),
+                 ("keyword.enabled", "high"),
+
+                 ("app.update.*", "high"),
+                 ("browser.addon-watch.*", "high"),
+                 ("datareporting.", "high"),
+                 ("extensions.blocklist.*", "high"),
+                 ("extensions.getAddons.*", "high"),
+                 ("extensions.update.*", "high"),
+
+                 ("security.*", "high"),
+
+                 ("network.proxy.*", "low"),
+                 ("network.http.*", "low"),
+                 ("network.websocket.*", "low"))
+
+        for pref, severity in PREFS:
+            yield test, pref, severity
+
+
+    def test_pref_composed_branches(self):
+        """
+        Tests that preference warnings still happen when branches are composed
+        via `getBranch`.
+        """
+
+        warnings = [
+            {"message": "Attempt to set a dangerous preference",
+             "signing_severity": "high"}]
+
+        self.run_script("""
+            Services.prefs.getBranch('browser.star')
+                    .setCharPref('tup.homepage', 'http://evil.com');
+        """)
+        self.assert_failed(with_warnings=warnings)
+
+
+    def test_profile_filenames(self):
+        """
+        Test that references to critical files in the user profile cause
+        warnings.
+        """
+
+        warnings = [
+            {"id": ("testcases_regex", "string", "profile_filenames"),
+             "message": "Reference to critical user profile data",
+             "signing_severity": "low"}]
+
+        def fail(script):
+            self.setUp()
+            self.run_script(script)
+            self.assert_failed(with_warnings=warnings)
+
+        paths = (r"addons.json",
+                 r"safebrowsing",
+                 r"safebrowsing\\foo.bar",
+                 r"safebrowsing/foo.bar")
+
+        patterns = ("'%s'",
+                    "'/%s'",
+                    "'\\%s'")
+
+        for path in paths:
+            for pattern in patterns:
+                yield fail, pattern % path
+
+        yield fail, "'addons' + '.json'"
+
+
+    def test_categories(self):
+        """Tests that dangerous category names are flagged in JS strings."""
+
+        warning = {"id": ("testcases_chromemanifest", "test_resourcemodules",
+                          "resource_modules"),
+                   "message": "Potentially dangerous category entry",
+                   "signing_severity": "medium",
+                   "editors_only": True}
+
+        self.run_script("'JavaScript-global-property'")
+        self.assert_failed(with_warnings=[warning])
+
+
+    def test_proxy_filter(self):
+        """Tests that registering a proxy filter generates a warning."""
+
+        warning = {"id": ("testcases_javascript_actions",
+                          "predefinedentities", "proxy_filter"),
+                   "signing_severity": "low",
+                   "editors_only": True}
+
+        self.run_script("""
+            Cc[""].getService(Ci.nsIProtocolProxyService)
+                 .registerFilter(foo, 0);
+        """)
+        self.assert_failed(with_warnings=[warning])
+
+
+    def test_addon_install(self):
+        """Tests attempts to install an add-on are flagged."""
+
+        warning = {"id": ("js", "traverser", "dangerous_global"),
+                   "editors_only": True,
+                   "signing_severity": "high"}
+
+        def test(method):
+            self.setUp()
+            self.run_script("""
+                AddonManager.%s(location, callback, plus, some, other, stuff);
+            """ % method)
+            self.assert_failed(with_warnings=[warning])
+
+        for method in (u"getInstallForFile",
+                       u"getInstallForURL"):
+            yield test, method
+
+
+    def test_addon_settings(self):
+        """Tests that attempts to change add-on settings via the
+        AddonManager API are flagged."""
+
+        warning = {
+            "description":
+                "Changing this preference may have severe security "
+                "implications, and is forbidden under most circumstances.",
+            "editors_only": True,
+            "signing_severity": "high"}
+
+        props = (u"autoUpdateDefault",
+                 u"checkUpdateSecurity",
+                 u"checkUpdateSecurityDefault",
+                 u"updateEnabled")
+
+        def test(prop):
+            self.setUp()
+            self.run_script("AddonManager.%s = false;" % prop)
+            self.assert_failed(with_warnings=[warning])
+
+        for prop in props:
+            yield test, prop
+
+
+    def test_ctypes(self):
+        """Tests that usage of `ctypes` generates warnings."""
+
+        self.run_script("""
+            ctypes.open("libc.so.6");
+        """)
+        self.assert_failed(with_warnings=[
+            {"id": ("js", "traverser", "dangerous_global"),
+             "editors_only": True,
+             "signing_severity": "high"}])
+
+
+    def test_nsIProcess(self):
+        """Tests that usage of `nsIProcess` generates warnings."""
+
+        self.run_script("""
+            Cc[""].createInstance(Ci.nsIProcess);
+        """)
+        self.assert_failed(with_warnings=[
+            {"id": ("js", "traverser", "dangerous_global"),
+             "editors_only": True,
+             "signing_severity": "high"}])
+
+
+    def test_eval(self):
+        """Tests that usage of eval-related functions generates warnings."""
+
+        functions = ("eval",
+                     "Function",
+                     "setTimeout",
+                     "setInterval")
+
+        warning = {"id": ("javascript", "dangerous_global", "eval"),
+                   "signing_severity": "high"}
+
+        def test(func):
+            self.setUp()
+            self.run_script("%s('doEvilStuff()')" % func)
+            self.assert_failed(with_warnings=[warning])
+
+        for func in functions:
+            yield test, func
+
+
+    def test_cert_service(self):
+        """Tests that changes to certificate trust leads to warnings."""
+
+        interfaces = ("nsIX509CertDB",
+                      "nsIX509CertDB2",
+                      "nsIX509CertList",
+                      "nsICertOverrideService")
+
+        contracts = ("@mozilla.org/security/x509certdb;1",
+                     "@mozilla.org/security/x509certlist;1",
+                     "@mozilla.org/security/certoverride;1")
+
+        warning = {"id": ("javascript", "predefinedentities", "cert_db"),
+                   "editors_only": True,
+                   "signing_severity": "high"}
+
+        def fail(script):
+            self.setUp()
+            self.run_script(script)
+            self.assert_failed(with_warnings=[warning])
+
+        for interface in interfaces:
+            yield fail, "Cc[''].getService(Ci.%s)" % interface
+
+        for contract in contracts:
+            yield fail, "Cc['%s'].getService()" % contract
+
+
+    def test_new_tab_page(self):
+        """Tests that attempts to replace about:newtab are flagged."""
+
+        patterns = (
+            "if (foo == 'about:newtab') doStuff();",
+            'if (bar === "about:blank") doStuff();',
+            "if (baz==='about:newtab') doStuff();",
+            "if ('about:newtab' == thing) doStuff();",
+            "/^about:newtab$/.test(thing)",
+            "/about:newtab/.test(thing)",
+            "'@mozilla.org/network/protocol/about;1?what=newtab'")
+
+        warning = {"signing_severity": "low"}
+
+        def fail(script):
+            self.setUp()
+            self.run_js_regex(script)
+            self.assert_failed(with_warnings=[warning])
+
+        for pattern in patterns:
+            yield fail, pattern
+
+
+    def test_script_creation(self):
+        """Tests that creation of script tags generates warnings."""
+
+        warning = {"id": ("testcases_javascript_instanceactions",
+                          "_call_expression", "called_createelement"),
+                   "signing_severity": "medium"}
+
+        self.run_script("""
+            doc.createElement("script");
+        """)
+        self.assert_failed(with_warnings=[warning])
+
+
+    def test_event_attributes(self):
+        """Tests that creation of event handler attributes is flagged."""
+
+        warning = {"id": ("testcases_javascript_instanceactions",
+                          "setAttribute", "setting_on*"),
+                   "signing_severity": "medium"}
+
+        self.run_script("""
+            elem.setAttribute("onhover", "doStuff();" + with_stuff);
+        """)
+        self.assert_failed(with_warnings=[warning])
+
+
+    def test_event_attributes_innerhtml(self):
+        """Tests that creation of event handler attributes via innerHTML
+        assignment is flagged."""
+
+        warning = {"id": ("testcases_javascript_instancetypes",
+                          "set_innerHTML", "event_assignment"),
+                   "signing_severity": "medium"}
+
+        self.run_script("""
+            elem.innerHTML = '<a onhover="doEvilStuff()"></a>';
+        """)
+        self.assert_failed(with_warnings=[warning])
--- a/tests/test_js_xpcom.py
+++ b/tests/test_js_xpcom.py
@@ -221,32 +221,6 @@ def test_xpcom_shortcut_services_ww():
     """)
 
 
-def test_xpcom_nsibrowsersearchservice():
-    """
-    Test that currentEngine/defaultEngine are flagged when they are
-    overwritten.
-    """
-
-    assert not _do_test_raw("""
-    var foo = Cc["@mozilla.org/browser/search-service;1"]
-                   .getService(Components.interfaces.nsIBrowserSearchService);
-    alert(foo.currentEngine);
-    foo.bar = true;
-    """).failed()
-
-    assert not _do_test_raw("""
-    var foo = Cc["@mozilla.org/browser/search-service;1"]
-                   .getService(Components.interfaces.nsIBrowserSearchService);
-    foo.currentEngine = 123;
-    """).failed()
-
-    assert not _do_test_raw("""
-    var foo = Cc["@mozilla.org/browser/search-service;1"]
-                   .getService(Components.interfaces.nsIBrowserSearchService);
-    foo.defaultEngine = 123;
-    """).failed()
-
-
 def test_synchronous_sql():
     """Test that uses of synchronous SQL are flagged."""
 
--- a/validator/errorbundler.py
+++ b/validator/errorbundler.py
@@ -1,8 +1,6 @@
 import json
-import sys
 import types
 import uuid
-from functools import partial
 from StringIO import StringIO
 
 from outputhandlers.shellcolors import OutputHandler
@@ -46,6 +44,7 @@ class ErrorBundle(object):
         self.compat_summary = {"errors": 0,
                                "warnings": 0,
                                "notices": 0}
+        self.signing_summary = {s: 0 for s in self.SEVERITIES}
 
         self.ending_tier = 1
         self.tier = 1
@@ -74,6 +73,8 @@ class ErrorBundle(object):
 
         self.supported_versions = self.for_appversions = for_appversions
 
+    SEVERITIES = {"trivial", "low", "medium", "high"}
+
     def _message(type_, message_type):
         def wrap(self, *args, **kwargs):
             message = {
@@ -93,6 +94,13 @@ class ErrorBundle(object):
             }
             for field in ("tier", "for_appversions", "compatibility_type", ):
                 message[field] = kwargs.get(field)
+            if "signing_severity" in kwargs:
+                severity = kwargs["signing_severity"]
+
+                assert severity in self.SEVERITIES
+
+                self.signing_summary[severity] += 1
+                message["signing_severity"] = severity
 
             self._save_message(getattr(self, type_), type_, message,
                                context=kwargs.get("context"))
@@ -304,6 +312,7 @@ class ErrorBundle(object):
                   "notices": len(self.notices),
                   "message_tree": self.message_tree,
                   "compatibility_summary": self.compat_summary,
+                  "signing_summary": self.signing_summary,
                   "metadata": self.metadata}
 
         messages = output["messages"]
@@ -417,6 +426,11 @@ class ErrorBundle(object):
                 verbose_output.append(
                     self._flatten_list(message["description"]))
 
+            if message.get("signing_severity"):
+                verbose_output.append(
+                    ("\tAutomated signing severity: %s" %
+                     message["signing_severity"]))
+
             # Show the user what tier we're on
             verbose_output.append("\tTier:\t%d" % message["tier"])
 
@@ -439,7 +453,7 @@ class ErrorBundle(object):
             if message["column"] and message["column"] != 0:
                 verbose_output.append("\tColumn:\t%d" % message["column"])
 
-            if "context" in message and message["context"]:
+            if message.get("context"):
                 verbose_output.append("\tContext:")
                 verbose_output.extend([("\t> %s" % x
                                         if x is not None
--- a/validator/testcases/installrdf.py
+++ b/validator/testcases/installrdf.py
@@ -53,7 +53,6 @@ def _test_rdf(err, install):
                       "updateInfoURL",
                       "updateKey",
                       "updateURL",
-                      "updateLink",  # Banned, but if not, pass it once.
                       "updateHash",
                       "signature",
                       "skinnable",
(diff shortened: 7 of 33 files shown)


tests/resources/installrdf/shouldnt_exist.rdf
```diff
@@ -8,6 +8,7 @@
 	<em:type>2</em:type>
 	
 	<em:updateURL>true</em:updateURL>
+	<em:updateKey>meh</em:updateKey>
```
magopian: Didn't know about this one ;) :+1:

validator/errorbundler.py
```diff
@@ -74,6 +72,8 @@ def __init__(self, determined=True, listed=True, instant=False,
 
         self.supported_versions = self.for_appversions = for_appversions
 
+    SEVERITIES = {"trivial", "low", "medium", "high"}
```
magopian: I don't remember seeing anything about `trivial` in your notes: I remember that a low severity would still let an addon be automatically reviewed (pass validation) if it had the exact same warning in a previously reviewed version. How about `trivial`?

validator/errorbundler.py
```diff
@@ -439,7 +447,7 @@ def _print_message(self, prefix, message, verbose=True):
             if message["column"] and message["column"] != 0:
                 verbose_output.append("\tColumn:\t%d" % message["column"])
 
-            if "context" in message and message["context"]:
+            if message.get("context"):
```
magopian: :+1:

validator/testcases/installrdf.py
```diff
@@ -53,7 +53,6 @@ def _test_rdf(err, install):
                       "updateInfoURL",
                       "updateKey",
                       "updateURL",
-                      "updateLink",  # Banned, but if not, pass it once.
```
magopian: Is that deprecated?

validator/testcases/installrdf.py
```diff
@@ -82,16 +81,19 @@ def _test_rdf(err, install):
 
     top_id = install.get_root_subject()
 
+    predicates = {}
+
     for pred_raw in install.rdf.predicates(top_id, None):
         predicate = pred_raw.split("#")[-1]
 
+        value = install.get_object(top_id, pred_raw)
```
magopian: nice factorization :+1:

validator/testcases/javascript/actions.py
```diff
@@ -81,16 +78,26 @@ def trace_member(traverser, node, instantiate=False):
...
                 traverser._debug("MEMBER_EXP>>XPCOM_WILDCARD")
+
+                from predefinedentities import CONTRACT_ENTITIES
+                if identifier in CONTRACT_ENTITIES:
+                    kw = dict(err_id=("js", "actions", "dangerous_contract"),
+                              warning="Dangerous XPCOM contract ID")
+                    kw.update(CONTRACT_ENTITIES[identifier])
+
+                    traverser.warning(**kw)
+
```
magopian: What does this change do? Is it related to the bug?

magopian: Ah, I see it's another test that you added a severity warning too, below.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -579,13 +579,30 @@
...
-                      u"nsIX509CertDB": {"dangerous": {
-                          "description": "Access to the X509 certificate "
-                                         "database is potentially dangerous "
-                                         "and requires careful review by an "
-                                         "administrative reviewer.",
-                          "editors_only": True,
-                      }}}
+                      }
+
+CONTRACT_ENTITIES = {}
```
magopian: This feels a bit weird, why not remove this line, and on line 600 do something like
`CONTRACT_ENTITIES = dict(....)`

magopian:
This looks promising! Do you plan on having a high level counter for the different severities, in the json returned? Similar to what's already done for errors/warnings, where it says how many there are without digging/looping on the messages themselves?

tests/resources/installrdf/shouldnt_exist.rdf
kmaglione: We just never checked for it before. I added it because we're now warning when it's missing, for unhosted add-ons.

validator/errorbundler.py
kmaglione: I haven't decided if I'm going to keep that yet. I added it for the moment because there was a warning that was signing-related, but I didn't want to trigger manual review. If it stays, the idea would be that it would show up as a signing-related message, but wouldn't prevent signing. Or maybe a bunch of them would, but a single one wouldn't.

validator/testcases/installrdf.py
kmaglione: It's never actually existed. It's an element in update.rdf, but not in install.rdf. Donno why it was there.

validator/testcases/javascript/predefinedentities.py
kmaglione: I went back and forth on that one. I mainly just wanted the `CONTRACT_ENTITIES` and `INTERFACE_ENTITIES` lines to follow the same pattern. But I'm happy to change it to just a `dict()` on that line.

kmaglione:
I hadn't planned on it, but it sounds like a good idea.

kmaglione:
Updated WIP. Addressed most comments and added some more checks. I made some non-trivial changes to the way the `"readonly"` property works (the old behavior had a lot of problems), so that might be worth giving some extra attention.

tests/helper.py
```diff
@@ -90,18 +90,29 @@ def assert_failed(self, with_errors=False, with_warnings=None):
...
+                       for prop, value in props.iteritems())
+
+        def find_message(messages, props):
+            return any(check_message(message, props)
+                       for message in messages)
+
+        def test_messages(mtype, expected):
+            messages = getattr(self.err, mtype)
+
+            if isinstance(expected, (list, tuple)):
```
magopian: This could be good enough for test code, but the proper way is to do something like http://stackoverflow.com/a/1952655

tests/helper.py
```diff
@@ -90,18 +90,29 @@ def assert_failed(self, with_errors=False, with_warnings=None):
...
+                assert all(find_message(messages, props)
+                           for props in expected)
+            elif expected:
```
magopian: Would that read better if it was written `elif expected is True:`?

tests/helper.py
```diff
@@ -90,18 +90,29 @@ def assert_failed(self, with_errors=False, with_warnings=None):
...
+                assert messages, "Expected %s." % mtype
+            elif expected is not None:
```
magopian: That would benefit from a comment like `if expected is False, or falsy like [], {}...`

tests/helper.py
```diff
@@ -90,18 +90,29 @@ def assert_failed(self, with_errors=False, with_warnings=None):
...
-            raise AssertionError("Tests found unexpected errors: %s" %
-                                self.err.print_summary(verbose=True))
-
-        if with_warnings is not None:
-            if with_warnings:
-                assert self.err.warnings, "Warnings were expected."
-            elif self.err.warnings:
-                raise ("Tests found unexpected warnings: %s" %
-                           self.err.print_summary())
+        def check_message(message, props):
```
magopian: Docstring or comment please

tests/helper.py
```diff
@@ -90,18 +90,29 @@ def assert_failed(self, with_errors=False, with_warnings=None):
...
+            return all(message.get(prop) == value
+                       for prop, value in props.iteritems())
+
+        def find_message(messages, props):
```
magopian: Docstring or comment please

tests/helper.py
magopian: Correction: if it's an iterable, it'll be treated by line 104. So the comment could be `if expected is False or 0`. It could also be rewritten `if expected in [False, 0]`?

Also, why not accept `None`? That would allow calling `test_messages('errors', None)` which is pretty clear? In that case the code would simply read `else:  # If expected is falsy.`.

tests/test_js_functions.py
```diff
@@ -57,6 +57,38 @@ def test_createElement():
...
+    """Tests that enablePrivilege throws a warning."""
+
+    err = _do_test_raw("""
+        netscape.security.PrivilegeManager.enablePrivilege()
+    """)
+    assert err.warnings
+
+    eq_(err.warnings[0]["id"],
+        ("js", "traverser", "dangerous_global"))
+    eq_(err.warnings[0].get("signing_severity"), "high")
```
magopian: why a `.get()` here?

tests/test_js_functions.py
```diff
@@ -57,6 +57,38 @@ def test_createElement():
...
+
+    for meth in "cloneInto", "exportFunction":
+        err = _do_test_raw("""
+            Cu.%s(foo)
+        """ % meth)
+        assert err.warnings
+
+        eq_(err.warnings[0]["id"],
+            ("js", "traverser", "dangerous_global"))
+        eq_(err.warnings[0].get("signing_severity"), "low")
```
magopian: Same here, why a `.get()`?

tests/test_js_functions.py
```diff
@@ -57,6 +57,38 @@ def test_createElement():
...
+    err = _do_test_raw("""
+        netscape.security.PrivilegeManager.enablePrivilege()
+    """)
+    assert err.warnings
+
+    eq_(err.warnings[0]["id"],
+        ("js", "traverser", "dangerous_global"))
+    eq_(err.warnings[0].get("signing_severity"), "high")
+
+def test_privileged_unprivileged_interacion():
```
magopian: typo `interaction`

tests/test_js_signing.py
```diff
@@ -0,0 +1,55 @@
+"""
+Various tests related to validation for automated signing.
+"""
+
+from js_helper import TestCase
```
magopian: `from .js_helper import TestCase`

tests/test_js_signing.py
```diff
@@ -0,0 +1,55 @@
...
+        def test(obj, prop, stuff, warnings):
+            self.setUp()
+            self.run_script("""%s.%s%s;""" % (obj, prop, stuff))
+            self.assert_failed(with_warnings=warnings)
+
+        objs = ("Cc[''].getService(Ci.nsIBrowserSearchService)",
+                "Services.search")
+
+        for obj in objs:
+            warnings = {"signing_severity": "high",
```
magopian: This can be declared out of the loop, one line above

tests/test_js_signing.py
```diff
@@ -0,0 +1,55 @@
...
+                        "id": ("testcases_javascript_actions",
+                               "search_service",
+                               "changes")},
+
+            for prop in "currentEngine", "defaultEngine":
+                yield test, obj, prop, " = foo", warnings
```
magopian: What does that do? It looks like some kind of hackery to "generate" tests for unittest/nose? I don't like it, I'm sure you can do the same with a simple call to your `test()` helper above.

To be honest, I'd rather have two separate tests here, and no loop, no helper.

tests/test_js_signing.py
```diff
@@ -0,0 +1,55 @@
...
+        def test(method):
+            self.setUp()
+            self.run_script("""
+                Cc[""].createInstance(Ci.nsIWindowsRegKey).%s(foo, bar);
+            """ % method)
+            self.assert_failed(with_warnings=warnings)
+
+        for method in ("create", "createChild", "writeBinaryValue",
+                       "writeInt64Value", "writeIntValue", "writeStringValue"):
+            yield test, method
```
magopian: you could get rid of the `yield` hack and paste the content of the `test()` helper in the loop.

tests/test_js_xpcom.py
```diff
@@ -221,32 +221,6 @@ def test_xpcom_shortcut_services_ww():
...
-    foo.currentEngine = 123;
-    """).failed()
-
-    assert not _do_test_raw("""
-    var foo = Cc["@mozilla.org/browser/search-service;1"]
-                   .getService(Components.interfaces.nsIBrowserSearchService);
-    foo.defaultEngine = 123;
-    """).failed()
-
-
```
magopian: Is that linked to this PR? If not, please split those in a separate PR.

validator/testcases/javascript/actions.py
```diff
@@ -81,16 +78,26 @@ def trace_member(traverser, node, instantiate=False):
...
 
+        identifier = _get_member_exp_property(traverser, node)
+
         # Handle the various global entity properties.
         if base.is_global:
             # If we've got an XPCOM wildcard, return a copy of the entity.
             if "xpcom_wildcard" in base.value:
                 traverser._debug("MEMBER_EXP>>XPCOM_WILDCARD")
+
+                from predefinedentities import CONTRACT_ENTITIES
```
magopian: Why not put that at the top of the module? Import loop? If so, please comment

validator/testcases/javascript/actions.py
```diff
@@ -81,16 +78,26 @@ def trace_member(traverser, node, instantiate=False):
...
+                if identifier in CONTRACT_ENTITIES:
+                    kw = dict(err_id=("js", "actions", "dangerous_contract"),
+                              warning="Dangerous XPCOM contract ID")
```
magopian: Why not define all the dict items in the same place (either here or in `prefedinedentities`)?

validator/testcases/javascript/actions.py
```diff
@@ -595,7 +602,9 @@ def _expr_assignment(traverser, node):
     if node["operator"] == "=":
 
         global_overwrite = False
-        readonly_value = True
+        # FIXME(Kris): This is not a great heuristic.
+        from predefinedentities import is_shared_scope
```
magopian: Same question as above, why not at the top of the module?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -64,6 +101,12 @@
 ]
 
 
+def is_shared_scope(t, r=None, rn=None):
```
magopian: Docstring and no one-letter variables please

validator/testcases/javascript/actions.py
```diff
@@ -595,7 +602,9 @@ def _expr_assignment(traverser, node):
...
+        readonly_value = is_shared_scope(traverser)
```
magopian: This is the non-trivial change to the `readonly` part you were talking about? @mstriemer could you please have a look?

validator/testcases/javascript/predefinedentities.py
mstriemer: I don't know what a "shared scope" is. Can you please explain that in the docstring?

validator/testcases/javascript/actions.py
```diff
@@ -643,33 +651,36 @@ def _expr_assignment(traverser, node):
...
+                         global_overwrite)
+        traverser._debug("ASSIGNMENT:DIRECT:READONLY %r" %
+                         readonly_value)
 
-        if (global_overwrite and
-            not traverser.is_jsm and
-            not traverser.err.get_resource("em:bootstrap") and
-            readonly_value == True):
+        if callable(readonly_value):
+            # The readonly attribute supports a lambda function that accepts
```
magopian: It's missing the end of the sentence?

validator/testcases/javascript/actions.py
```diff
@@ -643,33 +651,36 @@ def _expr_assignment(traverser, node):
...
                 err_id=("testcases_javascript_actions",
                         "_expr_assignment",
                         "global_overwrite"),
                 warning="Global variable overwrite",
                 description="An attempt was made to overwrite a global "
-                            "variable in some JavaScript code.",
-                filename=traverser.filename,
-                line=traverser.line,
-                column=traverser.position,
-                context=traverser.context)
```
magopian: We don't need those anymore?

validator/testcases/javascript/actions.py
```diff
@@ -939,7 +949,7 @@ def _get_as_num(value):
             return value
         else:
             return int(value)
-    except ValueError, TypeError:
+    except (ValueError, TypeError):
```
magopian: :+1:

validator/testcases/javascript/predefinedentities.py
```diff
@@ -11,52 +12,88 @@
 
 # A list of identifiers and member values that may not be used.
 BANNED_IDENTIFIERS = {
-    u"newThread": "Creating threads from JavaScript is a common cause "
-                  "of crashes and is unsupported in recent versions of the platform",
-    u"processNextEvent": "Spinning the event loop with processNextEvent is a common "
-                         "cause of deadlocks, crashes, and other errors due to "
-                         "unintended reentrancy. Please use asynchronous callbacks "
-                         "instead wherever possible",
+    u"newThread": (
```
magopian: You don't need parens here to continue the strings.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -11,52 +12,88 @@
...
-    "Extensions must not alter user preferences such as the current home "
-    "page, new tab page, or search engine, without explicit user consent, "
-    "in which a user takes a non-default action. Such changes must also "
-    "be reverted when the extension is disabled or uninstalled.")
+CUSTOMIZATION_PREF_MESSAGE = {
+    "description": (
+        "Extensions must not alter user preferences such as the current home "
+        "page, new tab page, or search engine, without explicit user consent, "
+        "in which a user takes a non-default action. Such changes must also "
+        "be reverted when the extension is disabled or uninstalled.",
```
magopian: Did you mean to have a tuple here?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -11,52 +12,88 @@
...
+            "must make proxy changes using other mechanisms. Installing a "
+            "proxy filter is the recommended alternative: "
+            "https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/"
+            "Reference/Interface/nsIProtocolProxyService#registerFilter()"),
+        "signing_severity": "low"}),
+    (u"network.proxy.", NETWORK_PREF_MESSAGE),
+    (u"network.http.", NETWORK_PREF_MESSAGE),
+    (u"network.websocket.", NETWORK_PREF_MESSAGE),
     (u"nglayout.debug.disable_xul_cache", None),
+    # Let's see if we can get away with this...
```
magopian: What do you mean?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -88,7 +131,59 @@
...
+SEARCH_MESSAGE = "Potentially dangerous use of the search service"
+SEARCH_DESCRIPTION = (
+    "Changes to the default and currently-selected search engine settings "
+    "may only take place after users have explicitly opted-in, by taking "
+    "a non-default action. Any such changes must be reverted when the add-on "
+    "making them is disabled or uninstalled.")
+
+def search_warning(severity="medium", editors_only=False,
+                   message=SEARCH_MESSAGE,
+                   description=SEARCH_DESCRIPTION):
```
magopian: You aren't using any other parameters than `severity` in your calls, do we still need those extra ones here? Are you planning on using them in this PR?

I'd advise against premature optimization.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -575,32 +689,72 @@
...
         return wrap
-    INTERFACE_ENTITIES[interface] = {"xpcom_map": construct(interface)}
+    if interface not in INTERFACE_ENTITIES:
+        INTERFACE_ENTITIES[interface] = {}
+    INTERFACE_ENTITIES[interface]["xpcom_map"] = construct(interface)
 
 
 def build_quick_xpcom(method, interface, traverser, wrapper=False):
     """A shortcut to quickly build XPCOM objects on the fly."""
+    extra = ()
```
magopian: Seems like it got quite a few changes, can you please document what it's doing now? Constructing "deep" objects?

magopian: I didn't see a test for those changes, did I miss it?

tests/helper.py
kmaglione: On May 5, 2015 6:47 AM, "Mathieu Agopian" <notifications@github.com> wrote:
>
> In tests/helper.py:
>
> > -                assert self.err.warnings, "Warnings were expected."
> > -            elif self.err.warnings:
> > -                raise ("Tests found unexpected warnings: %s" %
> > -                           self.err.print_summary())
> > +        def check_message(message, props):
> > +            return all(message.get(prop) == value
> > +                       for prop, value in props.iteritems())
> > +
> > +        def find_message(messages, props):
> > +            return any(check_message(message, props)
> > +                       for message in messages)
> > +
> > +        def test_messages(mtype, expected):
> > +            messages = getattr(self.err, mtype)
> > +
> > +            if isinstance(expected, (list, tuple)):
>
> This could be good enough for test code, but the proper way is to do
something like http://stackoverflow.com/a/1952655

I'd already changed this locally :)

tests/helper.py
kmaglione: On Tue, May 05, 2015 at 06:48:58AM -0700, Mathieu Agopian wrote:
>Would that read better if it was written `elif expected is True:`?

Maybe, but it generally feels like a bad practice to treat 
truthy values that aren't `True` as false.

tests/test_js_functions.py
kmaglione: On Tue, May 05, 2015 at 07:07:48AM -0700, Mathieu Agopian wrote:
>> @@ -57,6 +57,38 @@ def test_createElement():
>>      document.createElement();
>>      """)
>>
>> +def test_enablePrivilege():
>> +    """Tests that enablePrivilege throws a warning."""
>> +
>> +    err = _do_test_raw("""
>> +        netscape.security.PrivilegeManager.enablePrivilege()
>> +    """)
>> +    assert err.warnings
>> +
>> +    eq_(err.warnings[0]["id"],
>> +        ("js", "traverser", "dangerous_global"))
>> +    eq_(err.warnings[0].get("signing_severity"), "high")
>
>why a `.get()` here?

Not all messages have a `"signing_severity"` property. I guess 
the `KeyError` would make the test fail the same as the 
`AssertionError`, I just tend to avoid letting tests fail that 
way.

tests/test_js_functions.py
kmaglione: On Tue, May 05, 2015 at 07:10:11AM -0700, Mathieu Agopian wrote:
>Did you mean to make it a `low` signing severity?

Yes. I think this is relatively benign, but I could be convinced 
otherwise.

tests/test_js_signing.py
kmaglione: On Tue, May 05, 2015 at 07:15:07AM -0700, Mathieu Agopian wrote:
>> @@ -0,0 +1,55 @@
>> +"""
>> +Various tests related to validation for automated signing.
>> +"""
>> +
>> +from js_helper import TestCase
>
>`from .js_helper import TestCase`

I copied this from a different testcase. Validator hasn't used 
new-style relative imports in the past, but I suppose now that 
we're on 2.7 everywhere...

tests/test_js_signing.py
kmaglione: On Tue, May 05, 2015 at 07:20:43AM -0700, Mathieu Agopian wrote:
>This can be declared out of the loop, one line above

One of the members gets modified a few lines down. If I moved it 
out of the for loop, I'd have to revert those changes at the 
start of each iteration.

tests/test_js_signing.py
kmaglione: On Tue, May 05, 2015 at 07:24:06AM -0700, Mathieu Agopian wrote:
>What does that do? It looks like some kind of hackery to "generate" tests for unittest/nose? I don't like it, I'm sure you can do the same with a simple call to your `test()` helper above.
>
>To be honest, I'd rather have two separate tests here, and no loop, no helper.

I'm just following the same pattern as the more-recently-added 
tests. I think the idea is that multiple values can be tested 
without one failure short-circuiting the rest.

I don't really have a strong opinion, though. I was just trying 
to be consistent with newer code.

tests/test_js_signing.py
kmaglione: On Tue, May 05, 2015 at 07:31:25AM -0700, Mathieu Agopian wrote:
>you could get rid of the `yield` hack and paste the content of the `test()` helper in the loop.

See above. The `yield` hack causes tests to continue even after 
one fails. I'm not even sure how much of a hack it is. It's a 
feature of nose.

tests/test_js_xpcom.py
kmaglione: On Tue, May 05, 2015 at 07:32:32AM -0700, Mathieu Agopian wrote:
>Is that linked to this PR? If not, please split those in a separate PR.

Yes. This tests exactly the opposite of what it's supposed to be 
testing, so one of my new checks causes it to fail.

validator/testcases/javascript/actions.py
kmaglione: On Tue, May 05, 2015 at 07:42:16AM -0700, Mathieu Agopian wrote:
>>          # Handle the various global entity properties.
>>          if base.is_global:
>>              # If we've got an XPCOM wildcard, return a copy of the entity.
>>              if "xpcom_wildcard" in base.value:
>>                  traverser._debug("MEMBER_EXP>>XPCOM_WILDCARD")
>> +
>> +                from predefinedentities import CONTRACT_ENTITIES
>
>Why not put that at the top of the module? Import loop? If so, please comment

I'm not sure. There are 6 other local imports of 
`predefinedentities` in this file, and I didn't have the time to 
try to find any side-effects of making them global when I was 
writing this, so I opted to stay on the safe side.

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 07:49:35AM -0700, Mathieu Agopian wrote:
>> +def is_shared_scope(t, r=None, rn=None):
>
>Docstring and no one-letter variables please

I'd like to, but unfortunately these are passed as keyword args, 
so changing them requires massive changes elsewhere.

-- 
Kris Maglione

The trouble with the world is that the stupid are cocksure and the
intelligent full of doubt.
	--Bertrand Russell

validator/testcases/javascript/actions.py
kmaglione: On Tue, May 05, 2015 at 07:58:18AM -0700, Mathieu Agopian wrote:
>>
>> -        if (global_overwrite and
>> -            not traverser.is_jsm and
>> -            not traverser.err.get_resource("em:bootstrap") and
>> -            readonly_value == True):
>> +        if callable(readonly_value):
>> +            # The readonly attribute supports a lambda function that accepts
>
>It's missing the end of the sentence?

That it is. Unfortunately, I'm not entirely sure what it's 
supposed to be. I didn't actually change this line, just the 
code around it.

validator/testcases/javascript/actions.py
kmaglione: On Tue, May 05, 2015 at 08:00:20AM -0700, Mathieu Agopian wrote:
>>                  err_id=("testcases_javascript_actions",
>>                          "_expr_assignment",
>>                          "global_overwrite"),
>>                  warning="Global variable overwrite",
>>                  description="An attempt was made to overwrite a global "
>> -                            "variable in some JavaScript code.",
>> -                filename=traverser.filename,
>> -                line=traverser.line,
>> -                column=traverser.position,
>> -                context=traverser.context)
>
>We don't need those anymore?

No. mstriemer added these utility functions to `Traverser` a 
while ago to take care of those parameters, and has been slowly 
fixing legacy calls since.

validator/testcases/javascript/actions.py
kmaglione: On Tue, May 05, 2015 at 08:01:30AM -0700, Mathieu Agopian wrote:
>> -    except ValueError, TypeError:
>> +    except (ValueError, TypeError):
>
>:+1:

One of the benefits of skimming through the copious output of 
`flake8`...

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 08:07:15AM -0700, Mathieu Agopian wrote:
>Did you mean to have a tuple here?

Yes. Tuple descriptions are rendered as one paragraph per 
element.

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 08:09:14AM -0700, Mathieu Agopian wrote:
>>      (u"nglayout.debug.disable_xul_cache", None),
>> +    # Let's see if we can get away with this...
>
>What do you mean?

I'm slightly worried that this substring may come up in 
too many unrelated contexts, but I'd prefer not to make it more 
specific, since basically any preference in that branch should 
be flagged.

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 08:14:14AM -0700, Mathieu Agopian wrote:
>You aren't using any other parameters than `severity` in your calls, do we still need those extra ones here? Are you planning on using them in this PR?

Initially I was planning on using more of them, yes, and I still 
may before I'm done. I'll simplify if it turns out they're not 
needed.

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 08:59:39AM -0700, Mathieu Agopian wrote:
>>
>>
>>  def build_quick_xpcom(method, interface, traverser, wrapper=False):
>>      """A shortcut to quickly build XPCOM objects on the fly."""
>> +    extra = ()
>
>Seems like it got quite a few changes, can you please document what it's doing now? Constructing "deep" objects?

Essentially, I need these objects to expose multiple interfaces. 
There's support elsewhere for doing this when JavaScript 
requests them, e.g.,

```js
prefService.QueryInterface(Ci.nsIPrefBranch).QueryInterface(Ci.nsIPrefService);
```

But no way from doing so from Python that isn't a chore. I made 
several attempts to do this cleanly and without any duplication, 
but too much of the related code is such a mess that I didn't 
have the time. The logic in the `for iface in extra` loop is 
copied and simplified from `xpcom_constructor`.

I started working on another PR for jQuery validation a while 
ago that should move some of that logic to somewhere more 
accessible, and in the end simplify both of these functions.

kmaglione: On Tue, May 05, 2015 at 09:00:04AM -0700, Mathieu Agopian wrote:
>>
>>
>>  def build_quick_xpcom(method, interface, traverser, wrapper=False):
>>      """A shortcut to quickly build XPCOM objects on the fly."""
>> +    extra = ()
>
>I didn't see a test for those changes, did I miss it?

There really aren't any tests for most of the validator 
internals. I'm not even sure where I'd put it. I did add tests 
for the services in `Services` object (for which I implemented 
this), which check that the behavior is the same whether the 
services is instantiated in JS (as above), or accessed via the 
services object, so this is being tested indirectly.

validator/testcases/javascript/actions.py
kmaglione: On Tue, May 05, 2015 at 07:46:49AM -0700, Mathieu Agopian wrote:
>>          # Handle the various global entity properties.
>>          if base.is_global:
>>              # If we've got an XPCOM wildcard, return a copy of the entity.
>>              if "xpcom_wildcard" in base.value:
>>                  traverser._debug("MEMBER_EXP>>XPCOM_WILDCARD")
>> +
>> +                from predefinedentities import CONTRACT_ENTITIES
>> +                if identifier in CONTRACT_ENTITIES:
>> +                    kw = dict(err_id=("js", "actions", "dangerous_contract"),
>> +                              warning="Dangerous XPCOM contract ID")
>
>Why not define all the dict items in the same place (either here or in `prefedinedentities`)?

Well, the main reason is that I don't want to have to have to 
duplicate keys.

At the moment, all of the contract IDs in that dict have the 
same properties, so that's not an issue. However, it's always 
worried me that we tend not to do checks for contract IDs, only 
interface names, so I intend to add more in the future.

The slightly more complicated reason is that most of the 
definitions in `predefinedentities` follow a similar pattern, 
where they can accept a string, for only the description, a 
dict, which can define other properties as well, or a callable 
which... has completely different behavior depending on exactly 
where you use it.

All of this is extremely messy and inconsistent, so what I'd 
really like to do is unify the handling of all of them via some 
kind of helper function, which means any new definitions along 
these lines need to follow the same behavior in any case.

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 07:57:46AM -0700, Mark Striemer wrote:
>I don't know what a "shared scope" is. Can you please explain that in the docstring?

I'll add a docstring, but essentially:

Most of these properties are flagged because when add-ons load 
code into a shared scope, such as a window, or web page, and 
replace or modify them, they cause conflicts with other code. 
This heuristic is the same as the previous version of the code, 
and assumes (incorrectly) that bootstrapped add-ons will never 
load code into a shared scope, and otherwise only recognizes 
JavaScript modules as private scopes.

The main change in this code is that rather than assume that 
`"readonly"` in general only applies to shared scopes, it now 
needs to be explicitly specified.

tests/helper.py
kmaglione: On Tue, May 05, 2015 at 07:05:49AM -0700, Mathieu Agopian wrote:
>> +            return all(message.get(prop) == value
>> +                       for prop, value in props.iteritems())
>> +
>> +        def find_message(messages, props):
>> +            return any(check_message(message, props)
>> +                       for message in messages)
>> +
>> +        def test_messages(mtype, expected):
>> +            messages = getattr(self.err, mtype)
>> +
>> +            if isinstance(expected, (list, tuple)):
>> +                assert all(find_message(messages, props)
>> +                           for props in expected)
>> +            elif expected:
>> +                assert messages, "Expected %s." % mtype
>> +            elif expected is not None:
>
>Correction: if it's an iterable, it'll be treated by line 104. So the comment could be `if expected is False or 0`. It could also be rewritten `if expected in [False, 0]`?
>
>Also, why not accept `None`? That would allow calling `test_messages('errors', None)` which is pretty clear? In that case the code would simply read `else:  # If expected is falsy.`.

See the docstring. It took me a while to decipher the behavior 
of the old version, even with its docstring, but essentially 
`None` means we don't care whether there are warnings/errors, 
while `False` means we explicitly expect there to be none.

kmaglione: On Tue, May 05, 2015 at 07:05:49AM -0700, Mathieu Agopian wrote:
>Also, why not accept `None`? That would allow calling `test_messages('errors', None)` which is pretty clear? In that case the code would simply read `else:  # If expected is falsy.`.

Ah, sorry, I now see that I finished updating that docstring 
after I last pushed. I thought I'd done it before. For 
reference, then:

```python
        """
        Asserts that the error bundle has registered a failure. If
        `with_warnings` is any value any true value, or `None`, a warning
        is considered a failure.

        `with_warnings` or `with_errors` may be any of the following:

        * True: Messages of this type must be present.
        * False: Messages of this type must not be present.
        * None: Messages of this type may or may not be present.
        * Iterable of dicts: For dict returned by the iterator, at least
        one message must have a matching item for every key/value pair in the
        dict.
        """
```

(Note that this behavior, with relation to `True`/`False`/`None` 
is unchanged. I just made the behavior consistent between 
warnings and errors, and allowed specifying specific messages 
which were expected.)

validator/testcases/javascript/predefinedentities.py
kmaglione: On Tue, May 05, 2015 at 08:04:47AM -0700, Mathieu Agopian wrote:
>You don't need parens here to continue the strings.

I've added parens to most of these descriptions in the past, 
since they can optionally be tuples of strings rather than 
single strings, and it seemed sensible to use parens for all of 
them. I guess I don't really have an objection to removing them, 
though.

tests/helper.py
magopian: Where is this docstring? There's a typo (`is any value any true value`)

tests/test_js_functions.py
magopian: Makes sense

tests/test_js_signing.py
magopian: Yeah, if we have opportunities to improve the code we're copying/modifying anyways... let's take them ;)

tests/test_js_signing.py
magopian: We don't care if it fails straight away or continue, right? If the suite fails, we still have to fix, so I'd rather not have code like that.

validator/testcases/javascript/predefinedentities.py
magopian: Is that clearer for you @mstriemer ? I'm still feeling confused about all this, but I don't have any context, and don't know the amo-validator code.

tests/helper.py
kmaglione: It replaces the docstring for `assert_failed`, but like I said, I added it just after I pushed. It will be in my next update later today.

tests/test_js_signing.py
kmaglione: Yes, but if it fails and short-circuits the rest of the tests, it's not always clear which parts are failing, which often means multiple runs after partial fixes to see what's going on. This is particularly true for the other test cases that check both `Services.*` and `.getService(*)`, where the failure could be in the definition on the `Services` object or in the definition of the interface. If all of the tests run regardless of one failure, it's much more obvious where the problem lies before you start digging into code.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -121,8 +216,18 @@
          "value": {}},
     u"nsIBrowserSearchService":
         {"value":
-            {u"currentEngine": {"readonly": True},
-             u"defaultEngine": {"readonly": True}}},
+            {u"currentEngine":
+                {"readonly": search_warning(severity="high")},
```
mstriemer: Shouldn't `readonly` be `True` or `False`? Should the keys here be `"dangerous"` instead?

validator/testcases/javascript/predefinedentities.py
mstriemer: This seems like it is moving the conditional from `is_shared_scope() and readonly_value` to when `readonly_value` is set which probably makes sense. I don't see `readonly` being set to anything other than this though but maybe I'm missing something?

validator/testcases/javascript/predefinedentities.py
kmaglione: `"dangerous"` applies to any access, or in the case of functions, to calls. `"readonly"` only applies to attempts to write to the value, which is what's important for these cases.

As for `True` or `False`, `"readonly"` has accepted `True`/`False`/`lambda` for a long time. This patch changes it to also accept a message description or a `dict` of message properties the same way `"dangerous"` does.

validator/testcases/javascript/predefinedentities.py
kmaglione: It's set to a constant value in a lot of places. The only other place it's given a function is:

```py
    u"top": {"readonly": actions._readonly_top},
```

Though that case just emits a notice.

But that's only two instances, so I guess it shouldn't be a problem to change the argument names after all. I was thinking of `"dangerous"`, where lambdas are used much more often.

kmaglione:
Pushed updates in a second commit, in case that's easier to review. Will rebase into a single commit before merge, or sooner if you'd prefer.

This now has nearly all of the planned tests. There are a few which I haven't come up with a satisfactory way of testing (mainly because they're highly context-sensitive) and am considering pushing back until after the initial roll-out, or entirely into runtime testing.

magopian:
Huge work, impressive. What's missing?

I see there's a couple of `FIXME`s, should they be fixed before merging?

validator/testcases/regex.py
```diff
@@ -2732,3 +2736,113 @@ def tests(self):
...
+        "warning": "Reference to critical user profile data",
+        "description": "Critical files in the user profile should not be "
+                       "directly accessed by add-ons. In many cases, an "
+                       "equivalent API is available and should be used "
+                       "instead.",
+        "signing_severity": "low"}),
+    (DANGEROUS_CATEGORIES, DANGEROUS_CATEGORY_WARNING),
+)
+
+validate_string = RegexTest(STRING_REGEXPS).test
```
magopian: Given the complexity of the code, and the easiness of making mistakes with regular expressions, I think those should be tested (talking about munge_filename, PROFILE_REGEX, validate_string, and RegextTest).

kmaglione: On Mon, May 18, 2015 at 07:03:13AM -0700, Mathieu Agopian wrote:
>Given the complexity of the code, and the easiness of making mistakes with regular expressions, I think those should be tested (talking about munge_filename, PROFILE_REGEX, validate_string, and RegextTest).

You're probably right. I added tests for all of the permutations 
that these tests should match, but unit tests would definitely 
not hurt. Will do.

kmaglione:
On Mon, May 18, 2015 at 06:33:43AM -0700, Mathieu Agopian wrote:
>Huge work, impressive. What's missing?



>I see there's a couple of `FIXME`s, should they be fixed before merging?

I think all of the FIXMEs I added are relevant to issues in old 
code that I happened to stumble across, but not particularly 
relevant to these changes. Adding support to treat `instanceof` 
like `QueryInterface` would definitely help some of these tests, 
though.

-- 
Kris Maglione

For every complex problem there is an answer that is clear, simple,
and wrong.
	--H. L. Mencken

kmaglione:
On Mon, May 18, 2015 at 06:33:43AM -0700, Mathieu Agopian wrote:
>What's missing?

From my spec:

• Changes to #identity-box, #searchbar, or #search-container

Of these, I could check for references to #identity-box, but I 
can't check for sub-elements as easily, and I haven't found any 
example code making these kinds of changes, so I'm not entirely 
sure any checks I come up with would be worth it.

#search-container is, unfortunately, referenced so often, and in 
so may different ways, by add-ons trying to insert things next 
to it, that I haven't been able to think of any non-runtime 
checks that would be useful. #searchbar has similar problems to 
#identity-box. I know there are add-ons that modify it, but I 
don't have access to code that would give me a good idea of how.

• Add-on uninstall detection.

The API doesn't make this especially easy. The method is just 
named `uninstall`, and trying to figure out what objects may be 
add-on objects is just too much work without massive changes to 
the validator. In any case, this is more of a nice-to-have. 
Add-on installation is really the important case.

• `importScripts` with remote scripts.

I'm considering adding a check for static strings that are 
remote, but I mainly have the impression that this isn't used 
very often with static strings.

• `evalInSandbox` on sandboxes created with chrome principals.

After giving this some more thought, I think I'm going to add a 
check for any `evalInSandbox` usage. But trying to figure out if 
the sandbox was chrome-privileged turned out to be far too 
difficult.

• href="javascript:..."

Same as above, I'd mainly intended this to apply to chrome 
scopes, but I think I'll add some more general checks.

• Accessing anything outside of ProfD/TempD (how do we do this statically?)

I haven't yet come up with an answer to "how do we do this 
statically", aside from prohibiting `nsIFile` objects from being 
created with `initWithPath` when unnecessary, and then 
restricting what paths can be gotten from the directory service. 
But there are still other file access APIs, and implementing all 
of the necessary semantics would be extremely complicated.

-- 
Kris Maglione

I live in the world rather as a spectator of mankind than as one of
the species.
	--Joseph Addison

magopian:
So what are the next steps? Fixing the tests, and then releasing the current work?

kmaglione:
If you don't have any other comments, I'll write tests for the new regex test class and the checks I mentioned above (which should be trivial), and then I think we're good.

I think fixing the failing tests just involves rebasing onto master. They pass locally. All the failures here are from a missing spidermonkey executable.

kmaglione:
Alright, Travis tests fixed, commits squashed. Should be good to go.

magopian:
Thanks a lot for this huge work @kmaglione!

PR #286, opened 2015-05-28
Bug 1169183: Fix construction of multi-interface objects in Services …

kmaglione:
…global.


magopian:
Did you validate that it does fix the issue with the [XPI provided in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1169183#c4)?

If so, r+ ;)

kmaglione:
Yup.

PR #287, opened 2015-06-05
Bug 1172148: Propagate more message properties when merging sub-JARs.

kmaglione:


validator/errorbundler.py
```diff
@@ -280,18 +283,25 @@ def _merge_messages(self, messages, callback, name):
...
+                    "context",
+                    "description",
+                    "editors_only",
+                    "for_appversions",
+                    "line",
+                    "signing_severity",
+                    "tier")
+
+            kw = {key: message[key] for key in keys if key in message}
+
```
magopian: Why write it like that? I don't find it more readable to be honest. Also, you added `editors_only` and `signing_severity`, is that on purpose?.

magopian:
Nice, could you write a test for that?

r+wc

validator/errorbundler.py
kmaglione: The main thing is that with all the noise, it wasn't clear to me which keys were being propagated from the old messages, and which were missing/being changed in some other way.

I actually want to get rid of this merging code altogether and just store the filename stack when we push/pop a state, and take care of the filename modifications in `_save_message`, but I didn't want to do anything complicated for something that needed to go out quickly.

And, yes, `editors_only` and `signing_severity` are the two properties that were missing.

kmaglione:
Heh. I was afraid you'd ask for tests... I'll see if there are any tests that already use inner JARs. I might be able to do it with a reasonable amount of difficulty.

validator/errorbundler.py
kmaglione: Ah, right, and the other issue was that the arguments actually need to be left out if they're not in the original message. So I needed to switch to a dict of kwargs either way, and add conditionals for at least those two properties. Just handling all of the properties the same way seemed like the cleaner solution.

kmaglione:
Added test. Let me know if you want me to go back and change the way I'm handling copying over the cloned properties.

I'm intending to send a pull request that just gets rid of this code as explained in the comment as soon as possible, in any case, so I don't think it's worth the trouble for now.

tests/test_errorbundler.py
```diff
@@ -116,8 +116,11 @@ def test_states():
     bundle.warning((), "super nested warning")
     bundle.notice((), "super nested notice")
 
-    # Test that nested compatibility messages retain their value
-    bundle.notice("comp", "Compat Test notice", compatibility_type="error")
+    # Test that nested compatibility messages retain various properties
```
magopian: Missing "." at the end of the comment.

tests/test_errorbundler.py
```diff
@@ -153,6 +156,8 @@ def test_states():
 
         if message["id"] == "comp":
             assert message["compatibility_type"] == "error"
+            assert message["editors_only"] == True
+            assert message["signing_severity"] == "high"
```
magopian: I know you didn't add them yourself in this commit, but there's a couple of `print` statements in there. Would you mind removing them?

magopian:
Thanks for the test.

r+wc

kmaglione:
Updated tests.

magopian:
Excellent work, thanks! Let me know when you want that released and used on AMO

PR #290, opened 2015-06-30
fix typo: missing space in multiline string

myrdd:


magopian:
Excellent, thanks a lot @myrdd !

PR #291, opened 2015-07-01
install fastchardet from PyPi

myrdd:
When calling `setup.py`, fastchardet is not installed, because `-e` is excluded. However, fastchardet is available from PyPi, so I suggest to use that.

Another solution (or an additional one) is to remove the exception for `-e` in `setup.py`, see https://github.com/mozilla/amo-validator/blob/v1.7.0/setup.py#L18.


requirements.txt
```diff
@@ -4,5 +4,5 @@ Fabric==0.9.0
 nose==1.0.0
 rdflib==3.2.1
 argparse==1.1
--e git://github.com/mattbasta/fastchardet#egg=fastchardet
+fastchardet
```
magopian: Please pin the version number: `fastchardet==0.2.0` (which is the latest version on pypi, and seems to also be the last commit on `master`, the one we've been using with the `-e` option)

magopian:
Excellent catch! Thanks a lot for that, very useful.

r+wc

P.S: `r+wc` means
1/ `r+`: reviewed, and approved (in opposition to `r-`)
2/ `wc`: with comments (with some comments or fixes needed)
Answering the comments should mean that this PR is ready to be merged!

requirements.txt
myrdd: Thanks for the quick reply! I'll do that. However, do you use to amend the changes and/or squash?

magopian: one or the other, as you prefer. For big changes, creating a new commit is easier to review, and then we squash before merging. For such a small change, amending is fine ;)

myrdd: Okay, but since I didn't want to force-push, I created a new commit.

magopian: It's also good. Amending or squashing will always need a force push. I can do the squashing myself when merging, which i'm going to do know. Thanks!

magopian:
Merged in https://github.com/mozilla/amo-validator/commit/f3c6b9d60133453a61141f3c341609a4d0efc21c

magopian:
Thanks @myrdd !

PR #293, opened 2015-07-01
Bug 1179486: Fix errors with non-ASCII strings in preference files.

kmaglione:


magopian:
Is there a way to test that "for real"? I tried simply adding `pref("extensions.foo-bar.\u263a", true);` to an add-on, and running the validator on it (using the commandline), but it didn't fail (at least I don't think it did ;)

Is there a bug for that? If not, should there be one? Is it to fix one of the sentry errors?

kmaglione:
It is for one of the sentry errors, yes:

http://sentry.mktmon.services.phx1.mozilla.com/mkt/addonsmozillaorg/group/16250/

This change fixes the error in that add-on.

You need to add the string to a `defaults/preferences/*.js` file, and you need to actually add `pref("extensions.foo-bar.☺", true);`. I added the unescaped variant just to be sure.

I could make a good argument that there should be a bug filed. Let me know if you want me to file one and update the commit summary.

kmaglione:
Incidentally, it looks like that variant doesn't show an error because we do some strange preprocessing on the JS sources before passing them to the parser... Will file a bug for that too.

kmaglione:
Filed bug and updated summary.

magopian:
Using `pref("extensions.foo-bar.éç", true);` "works" to raise the exception, and indeed this PR fixes this.

magopian:
r+

tests/test_js_prefs.py
```diff
@@ -23,6 +23,17 @@ def test_pref_innocuous_branch():
...
+    """Tests that non-ASCII preferences do not cause errors."""
+
+    assert not _do_test_raw("""
+    pref("extensions.foo-bar.\u263a", true);
+    """, path="defaults/preferences/prefs.js").failed()
+
+    assert not _do_test_raw("""
+    pref("extensions.foo-bar.\\u263a", true);
+    """, path="defaults/preferences/prefs.js").failed()
+
```
magopian: Missing the test with `☺` and for good measure try `éç` which is the only one that raised the exception, at least on my computer (even with the patch from #294).

kmaglione: Hm. The first test just needs to be a `u` string. As it is now, the `\u263a` gets passed through literally. With that change, the first test fails prior to this fix. With #294, they both fail prior to this fix.

magopian:
Thanks! Don't forget to update the bug with the merge commit (i've done it for this one)

PR #294, opened 2015-07-01
Bug 1179490: Don't strip \\u and \\x escapes from scripts before parsing JS.

kmaglione:


magopian:
Do we have a clue why we had that in the first place?

magopian:
Missing a test

kmaglione:
I don't know why we had it, no. I tracked down an older revision with this comment:

```py
    code = unicodehelper.decode(code)
    # Acceptable unicode characters still need to be stripped. Just remove the
    # slash: a character is necessary to prevent bad identifier errors.
    return JS_ESCAPE.sub("u", code)
```

But it doesn't make sense, and everything works as expected (and all tests pass) without this mangling.

kmaglione:
Added test.

magopian:
r+, thanks!

wagnerand:
@magopian @kmaglione merge?

PR #296, opened 2015-07-02
Bug 1179554: Convert preference and extension manager regexp tests to static-string-only tests

kmaglione:


kmaglione:
Sorry, I had to include the commit that allows regex tests to operate on arbitrary files in order to fix a test. If you want, I can move that to a separate PR, and try to fix the test in a different way for now.

magopian:
Making it a separate commit (to allow regex tests to operate on arbitrary files) is a good thing, and I believe it's ok to be in this PR (I'm just not sure that the stuff with `hbs` has anything to do with it).

magopian:
Nice work, r+wc

PR #300, opened 2015-07-09
Bug 1163799: Fix issues with the celery queues when validator tasks fail

kmaglione:
This is part of a two-part fix for issues in the task queues when validator tasks fail.

The current code never cleans up its timers or signal handlers before returning, which means that the timer will fire after the timeout interval even after the `validate` function has returned. When the task completes successfully, this is relatively harmless, since it sets a flag that essentially disarms the signal handler callback. When there's an exception, the flag is not set, which means the timer fires at an arbitrary point during the execution of the task process, and raises a `ValidationTimeout` exception, likely breaking something important.

This code cleans up the timeout and signal handler after they're no longer needed, and installs a timeout by default only when launched from as a system command.

A companion pull request for Olympia will move the handling of timeouts to Celery, where it belongs.

### DIFF
--- a/validator/loader.py
+++ b/validator/loader.py
@@ -1,3 +1,12 @@
+"""
+Presumably, the purpose of this module is to load things in an order such
+that they do not cause import loops. Given how easy it is, at the moment,
+to cause import loops by making trivial changes, this presumption is likely
+valid.
+
+It probably also has the effect of making sure each of these modules registers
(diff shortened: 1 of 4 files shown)


magopian:
This is excellent, well done investigating/coming up with this fix, let's hope it does fix our issue of queues backing up!

PR #303, opened 2015-07-11
Bug 1182729: Flag use of `contentScript` properties with non-static strings as a high severity signing error.

kmaglione:


magopian:
r+wc

PR #304, opened 2015-07-16
constants attribute name fix

l-hedgehog:


kmaglione:
Huh. I see we don't have any tests that even touch this file.

Thanks for the fix. Can you add a simple test for this? I think it would suffice for now to create a `tests/test_main.py` file with something like:

```python
from validator.main import main

def test_main():
    """Test that `main()` initializes without errors."""
    try:
        main()
    except SystemExit:
        pass
```

l-hedgehog:
I just copied and pasted the whole thing :smile: thanks!

kmaglione:
Heh. Great. Thanks again!

PR #305, opened 2015-07-20
Bug 1183399: Add suggestions for addressing each validator message with a signing severity.

kmaglione:
These should be all of the validator-side changes. Olympia changes will be required to display the messages, but they won't cause any trouble on the Olympia side in the mean time.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,6 +1,11 @@
+import mock
+import sys
+
 from validator.main import main
 
-def test_main():
+
+@mock.patch.object(sys, 'stderr')
+def test_main(stderr):
     """Test that `main()` initializes without errors."""
     try:
         main()
--- a/validator/testcases/javascript/instanceactions.py
+++ b/validator/testcases/javascript/instanceactions.py
@@ -108,6 +108,11 @@ def _create_script_tag(traverser):
         description="Dynamic creation of script nodes can be unsafe if "
                     "contents are not static or are otherwise unsafe, "
                     "or if `src` is remote.",
+        signing_help="Please avoid using <script> tags to load scripts. "
+                     "For potential alternatives, please see "
+                     "https://developer.mozilla.org/en-US/Add-ons/"
+                     "Overlay_Extensions/XUL_School/"
+                     "Appendix_D:_Loading_Scripts",
         signing_severity="medium")
 
 
@@ -146,6 +151,11 @@ def setAttribute(args, traverser, node, wrapper):
             description="To prevent vulnerabilities, event handlers (like "
                         "'onclick' and 'onhover') should always be defined "
                         "using addEventListener.",
+            signing_help="Please use `addEventListener` any place you might "
+                         "otherwise create event listener attributes. Event "
+                         "listener attributes will not be accepted in add-ons "
+                         "submitted for automated signing in any instance "
+                         "where they may be reasonably avoided.",
             signing_severity="medium")
 
 
--- a/validator/testcases/javascript/instanceproperties.py
+++ b/validator/testcases/javascript/instanceproperties.py
@@ -29,6 +29,19 @@ def _set_HTML_property(function, new_value, traverser):
         if isinstance(literal_value, types.StringTypes):
             # Static string assignments
 
+            HELP = ("Please avoid including JavaScript fragments in "
+                    "HTML stored in JavaScript strings. Event listeners "
+                    "should be added via `addEventListener` after the HTML "
+                    "has been injected.",
+                    "Injecting <script> nodes should be avoided when at all "
+                    "possible. If you cannot avoid loading a script directly "
+                    "into a content document, please consider doing so via "
+                    "the subscript loader (http://mzl.la/1VGxOPC) instead. "
+                    "If the subscript loader is not available, then the "
+                    "script nodes should be created using `createElement`, "
+                    "and should use a `src` attribute pointing to a "
+                    "`resource:` URL within your extension.")
+
             # Test for on* attributes and script tags.
             if EVENT_ASSIGNMENT.search(literal_value.lower()):
                 traverser.warning(
@@ -41,6 +54,7 @@ def _set_HTML_property(function, new_value, traverser):
                                  % function,
                                  "Event handler code: %s"
                                  % literal_value.encode("ascii", "replace")),
+                    signing_help=HELP,
                     signing_severity="medium")
             elif ("<script" in literal_value or
                   JS_URL.search(literal_value)):
@@ -53,6 +67,7 @@ def _set_HTML_property(function, new_value, traverser):
                                 "pages via script tags or JavaScript URLs. "
                                 "Instead, use event listeners and external "
                                 "JavaScript." % function,
+                    signing_help=HELP,
                     signing_severity="medium")
             else:
                 # Everything checks out, but we still want to pass it through
@@ -96,6 +111,10 @@ def set_on_event(new_value, traverser):
                         "assigned by setting an on* property to a "
                         "string of JS code. Rather, consider using "
                         "addEventListener.",
+            signing_help="Please add event listeners using the "
+                         "`addEventListener` API. If the property you are "
+                         "assigning to is not an event listener, please "
+                         "consider renaming it, if at all possible.",
             signing_severity="medium")
     elif (not is_literal and isinstance(new_value.value, jstypes.JSObject) and
           "handleEvent" in new_value.value.data):
@@ -128,6 +147,14 @@ def set__exposedProps__(new_value, traverser):
             "scopes is dangerous, and has been deprecated. If objects "
             "must be exposed to unprivileged scopes, `cloneInto` or "
             "`exportFunction` should be used instead."),
+        signing_help="If you are using this API to expose APIs to content, "
+                     "please use `Components.utils.cloneInto`, or "
+                     "`Components.utils.exportFunction` "
+                     "(http://mzl.la/1fvvgm9). If you are using it "
+                     "for other purposes, please consider using a built-in "
+                     "message passing interface instead. Extensions which "
+                     "expose APIs to content will be required to go through "
+                     "manual code review for at least one submission.",
         signing_severity="high")
 
 
@@ -152,6 +179,8 @@ def set_contentScript(value, traverser):
                         "is dangerous and error-prone. Please use a separate "
                         "JavaScript file, along with the "
                         "`contentScriptFile` property instead.",
+            signing_help="Please do not use the `contentScript` property "
+                         "in any add-ons submitted for automated signing.",
             signing_severity="high")
 
 
--- a/validator/testcases/javascript/predefinedentities.py
+++ b/validator/testcases/javascript/predefinedentities.py
@@ -21,6 +21,12 @@ BANNED_IDENTIFIERS = {
         "Please use asynchronous callbacks instead wherever possible",
 }
 
+CUSTOMIZATION_API_HELP = (
+    "We are currently working to provide libraries and APIs to allow "
+    "extensions to modify these settings in ways that we can guarantee are "
+    "in-policy. In the interim, we recommend that you avoid changing these "
+    "settings altogether, if at all possible.")
+
 CUSTOMIZATION_PREF_MESSAGE = {
     "description": (
         "Extensions must not alter user preferences such as the current home "
@@ -29,6 +35,10 @@ CUSTOMIZATION_PREF_MESSAGE = {
         "be reverted when the extension is disabled or uninstalled.",
         "In nearly all cases, new values for these preferences should be "
         "set in the default preference branch, rather than the user branch."),
+    "signing_help":
+        "Add-ons which directly change these preferences must undergo at "
+        "manual code review for at least one submission. " +
+        CUSTOMIZATION_API_HELP,
     "signing_severity": "high",
 }
 
@@ -36,6 +46,11 @@ NETWORK_PREF_MESSAGE = {
     "description":
         "Changing network preferences may be dangerous, and often leads to "
         "performance costs.",
+    "signing_help":
+        "Changes to these preferences are strongly discouraged. If at all "
+        "possible, you should remove any reference to them from "
+        "your extension. Extensions which do modify these preferences "
+        "must undergo light manual code review for at least one submission.",
     "signing_severity": "low",
 }
 
@@ -45,6 +60,15 @@ SEARCH_PREF_MESSAGE = {
         "All such changes must be made only via the browser search service, "
         "and only after an explicit opt-in from the user. All such changes "
         "must be reverted when the extension is disabled or uninstalled.",
+    "signing_help": (
+        "You should remove all references to these preferences from your "
+        "code, and interact with search settings only via the "
+        "`Services.search` interface. Extensions which interact with these "
+        "preferences directly are not acceptable within the Firefox add-on "
+        "ecosystem.",
+        "Note, however, that extensions which change search settings even via "
+        "the search service must undergo manual code review for at least "
+        "one submission. " + CUSTOMIZATION_API_HELP),
     "signing_severity": "high",
 }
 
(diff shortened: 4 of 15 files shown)


tests/test_main.py
```diff
@@ -1,6 +1,11 @@
+import mock
+import sys
+
 from validator.main import main
 
-def test_main():
+
+@mock.patch.object(sys, 'stderr')
```
kmaglione: This isn't related to the PR, but this is the first time I noticed this problem, and I didn't want it in my test output.

magopian: Can you please explain what issues you had? I'd rather have a fix for the cause, not the consequences. I'm afraid this will trip us in the future, wondering why we don't get the output we're expecting?

kmaglione: Essentially, this just tests that `main` runs up until the point that `parse_args` quits due to missing required arguments. The issue is that it prints a usage message directly to stderr, which shows up in the test output.

We really need proper tests for `main.py`. This test is just acting as a stopgap to make sure that the file loads and initializes correctly.

validator/testcases/javascript/instanceactions.py
```diff
@@ -166,6 +171,11 @@ def setAttribute(args, traverser, node, wrapper):
             description="To prevent vulnerabilities, event handlers (like "
                         "'onclick' and 'onhover') should always be defined "
                         "using addEventListener.",
+            signing_help="Please use `addEventListener` any place you might "
+                         "otherwise create event listener attributes. Event "
+                         "listener attributes will not be accepted in add-ons "
+                         "submitted for automated signing in any instance "
+                         "where they may be reasonably avoided.",
```
magopian: Does that sentence imply that event listener attributes used in places where they may not be reasonably avoided (examples?) could get automatically signed?

validator/testcases/javascript/instanceproperties.py
```diff
@@ -188,6 +207,14 @@ def set__exposedProps__(new_value, traverser):
...
             "must be exposed to unprivileged scopes, `cloneInto` or "
             "`exportFunction` should be used instead."),
+        signing_help="If you are using this API to expose APIs to content, "
+                     "please use `Components.utils.cloneInto`, or "
+                     "`Components.utils.exportFunction`. If you are using it "
+                     "for other purposes, please consider using a built-in "
+                     "message passing interface instead. Extensions which "
+                     "expose APIs to content will be required to go through "
+                     "at least one round of manual code review prior to "
+                     "signing.",
```
magopian: I thought only "low" severity warnings would be ignored the second time around? This suggests that if it goes through one manual code review, it could get automatically signed the following times? Even though it's a `high` severity?

validator/testcases/javascript/instanceproperties.py
```diff
@@ -225,6 +252,8 @@ def set_contentScript(value, traverser):
                         "is dangerous and error-prone. Please use a separate "
                         "JavaScript file, along with the "
                         "`contentScriptFile` property instead.",
+            signing_help="Please do not use the `contentScript` property "
+                         "in any add-ons submitted for automated signing.",
```
magopian: That sounds a bit confusing, should it say something along the lines of `add-ons that use the contentScript property won't be automatically signed`?

tests/test_main.py
magopian: Oh, I remember now, yes indeed. Also, it's not a "base test that could be inherited/re-used" elsewhere, so my comment is irrelevant. Thanks.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -36,6 +45,11 @@
     "description":
         "Changing network preferences may be dangerous, and often leads to "
         "performance costs.",
+    "signing_help":
+        "Changes to these preferences are strongly discouraged. If at all "
+        "possible, you should remove any reference to them from "
+        "your extension. Extensions which do modify these preferences "
+        "must undergo one round of light manual review.",
```
magopian: + `at least`?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -45,6 +59,14 @@
         "All such changes must be made only via the browser search service, "
         "and only after an explicit opt-in from the user. All such changes "
         "must be reverted when the extension is disabled or uninstalled.",
+    "signing_help": (
+        "You should remove all references to these preferences from your "
+        "code, and interact with search settings only via the "
+        "`Services.search` interface. Extensions which interact with these "
+        "preferences directly will not be accepted for signing.",
```
magopian: Did you mean that they won't be accepted at all? In which case, I think it would be clearer to remove the end of the sentence `for signing`.

validator/testcases/javascript/predefinedentities.py
```diff
@@ -528,6 +577,9 @@ def registry_key(write=False):
                 "method should always return its third argument in cases "
                 "when it is not supplying a specific proxy."),
             "editors_only": True,
+            "signing_help": "Due to the potential for unintended effects, "
+                            "any add-on which uses this API must undergo "
+                            "at least one round of light manual review.",
```
magopian: `light`? Do we have light manual reviews?Are you referring to preliminary review here? Not sure it's relevant to developers, as they don't explicitly choose their review queue (it only depends on the side-loading or not of their add-ons).

magopian:
Test failures to fix.

There's a few places where the wording suggests that after a first round of manual review, the following submissions will be automatically signed. However, those warnings sometimes have severities above `low`, and I thought we would only ignore those.

I guess it'll depend on the UI you're talking about in https://bugzilla.mozilla.org/show_bug.cgi?id=1172044#c2, to allow editors to also mark warnings as `low`?

Some of the `signing_help` messages added would probably benefit from links to the MDN, have you thought about that?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -1085,6 +1187,11 @@ def interface_obj(iface):
                             {"value":
                                  {u"enablePrivilege":
                                       {"dangerous": {
+                                          "signing_help":
+                                            "Any references to this API must "
+                                            "be removed from your extension. "
+                                            "Add-ons using this API will not "
+                                            "be accepted for signing.",
```
magopian: Same as above, not sure `for signing` is useful here (and may even be confusing)

validator/testcases/javascript/predefinedentities.py
```diff
@@ -922,6 +1020,14 @@ def interface_obj(iface):
...
             "reviewed by a qualified reviewer."),
         "editors_only": True,
+        "signing_help": ("Please try to avoid interacting with or bundling "
+                         "native binaries whenever possible. If you are "
+                         "bundling binaries for performance reasons, please "
+                         "consider alternatives such as Emscripten, "
+                         "JavaScript typed arrays, and Worker threads.",
+                         "Any code which makes use of the `ctypes` API "
+                         "must undergo at least one round of manual code "
+                         "review."),
```
magopian: This seems to be the exact same message as https://github.com/mozilla/amo-validator/pull/305/files#diff-2398abc7b75c0fb4a06ed98de3770c0aR54, could it be reused?

validator/testcases/regex.py
```diff
@@ -257,6 +257,14 @@ def tests(self):
...
                 "If this code does not change the behavior of the new tab "
                 "page, it may be ignored."),
+             "signing_help":
+                "Extensions may not programmatically override the new tab "
+                "page. If this code has another purpose, we nonetheless "
+                "recommend against testing URLs for these values, since "
+                "results can be unpredictable, and better options usually "
+                "exist. If you cannot avoid making these tests, please leave "
+                "this code unchanged, and it will be ignored in future "
+                "submissions.",
```
magopian: I liked the other wording better (saying that it'll need at least one manual review)

validator/testcases/regex.py
```diff
@@ -2901,9 +2915,34 @@ def munge_filename(name):
...
-STRING_REGEXPS += PREF_REGEXPS
+# For tests in literal strings, add help text suggesting passing the
+# preference directly to reference getter functions.
+PREF_STRING_HELP = (
+    "If you are reading, but not writing, this preference, please consider "
+    "passing a string literal directly to `Preferences.get()` or "
+    "`nsIPrefBranch.get*Pref`.")
+
+
+def maybe_tuple(value):
```
magopian: How about naming this `to_tuple`? Also, wouldn't be against a docstring (even if the code is easy to understand)

validator/testcases/javascript/instanceactions.py
```diff
@@ -128,6 +128,11 @@ def _create_script_tag(traverser):
         description="Dynamic creation of script nodes can be unsafe if "
                     "contents are not static or are otherwise unsafe, "
                     "or if `src` is remote.",
+        signing_help="Please avoid using <script> tags to load scripts. "
+                     "For potential alternatives, please see "
+                     "https://developer.mozilla.org/en-US/Add-ons/"
```
mstriemer: I think we normally remove the `en-US` bit from MDN URLs.

validator/testcases/regex.py
```diff
@@ -2901,9 +2915,34 @@ def munge_filename(name):
...
+    if isinstance(value, tuple):
+        return value
+    return (value,)
```
mstriemer: I'd rather see this in an `else:`. It makes it clearer that there are two separate cases, in my opinion.

```py
def to_tuple(value):
    if isinstance(value, tuple):
        return value
    else:
        return (value,)

Comment (magopian) [inline validator/testcases/regex.py]: I disagree on this very topic, but I guess it's a matter of taste ;)

File: validator/testcases/regex.py [hunk 14]
```diff
@@ -2901,9 +2915,34 @@ def munge_filename(name):
...
+
+
+def add_pref_help(desc):
+    desc = desc.copy()
+    for key in "description", "signing_help":
+        if key in desc:
+            desc[key] = maybe_tuple(desc[key]) + maybe_tuple(PREF_STRING_HELP)
```
mstriemer: If `description` or `signing_help` isn't in the dictionary then this won't add `PREF_STRING_HELP` on its own, is that intended?

validator/testcases/regex.py
mstriemer: The `if` isn't some special early return, it is one of the two cases. If we handles a third case for `list` then we'd get:

```py
def to_tuple(value):
    if isinstance(value, tuple):
        return value
    elif isinstance(value, list):
        return tuple(*value)
    return (value,)
```

I don't know what makes that last case special, it's just the default so it goes in the `else`.

validator/testcases/javascript/predefinedentities.py
kmaglione: I thought about that. The last sentence is (importantly) different, so the first sentence could be reused. I mainly decided not to try because it's way to easy to trigger import loops in the validator at the moment.

validator/testcases/javascript/predefinedentities.py
kmaglione: Essentially, some warnings mean that an editor just has to glance at the code and then approve. Others require much more thorough investigation. This is of the former type, hence light review.

validator/testcases/javascript/predefinedentities.py
kmaglione: I guess I mean they won't be accepted at all, yeah. But "will not be accepted" doesn't feel right without a qualifier. I'll try to think of a different wording.

validator/testcases/javascript/predefinedentities.py
kmaglione: "At least the first submission of any add-on which modifies these preferences must undergo manual review."

validator/testcases/javascript/instanceproperties.py
kmaglione: We'll currently accept it in any add-ons submitted for manual code review (listed add-ons), just not add-ons submitted for automated signing. Which means that if those add-ons wind up in the review queue with this issue, they'll be rejected.

validator/testcases/javascript/instanceproperties.py
kmaglione: Editors will be able to flag warnings they consider relatively safe to be ignored in future updates, regardless of the severity. If a reasonably trustworthy add-on is using these APIs, it will probably get a pass.

validator/testcases/javascript/instanceactions.py
kmaglione: I can't think of any good examples off hand, and I think that they'll rarely get a pass, but I know I've come across usages in the past that I'd probably approve. The nearest example I can think of is copying event handler attributes from one element to another, within the same document.

validator/testcases/javascript/instanceactions.py
kmaglione: We have a mix, some with `/en-US/`, some without. I thought about removing it, but translations of pages like this tend to be sketchy/out-of-date, and given that the validator messages are not localized anyway, I'd prefer to link to the English version by default.

validator/testcases/regex.py
kmaglione: In the other cases, I want to encourage developers to consider dropping the "features" that cause the flags. Here, I'd prefer that developers not change their code just to avoid the warning, so I chose different wording that I thought made that clearer.

validator/testcases/regex.py
kmaglione: I went with `maybe_tuple` because that naming has become an idiom in a lot of places.

It was originally a local function, where its purpose was pretty clear. I forgot to add a docstring when I exported it for tests.

validator/testcases/regex.py
kmaglione: If we handled a third case for some other type, I'd write the second conditional as a second `if` rather than an `elif`. Only I'd actually probably just make it `isinstance(value, (list, tuple))`.

In any case, I don't really have a strong opinion here, but I generally lean toward the `else`-less form, since the `else` is superfluous, and doesn't really make anything clearer.

validator/testcases/regex.py
kmaglione: Yes. If it's already a signing-related message, it should already have a `signing_help` property. And if it isn't a signing-related message, I don't want to add that property.

validator/testcases/javascript/instanceproperties.py
magopian: So, maybe it should state that in the message? I personally did not know there was a difference in the review between listed and unlisted add-ons (apart from the "policy" restrictions). So, just to clarify: `contentScript` will never be allowed in unlisted add-ons? Or maybe only in side-loaded add-ons?

kmaglione: Side-loaded add-ons are reviewed in the same way as listed add-ons. Other unlisted add-ons are only reviewed if they don't pass automated validation, and then, only to verify that the warnings are safe (unless they happen to look suspicious and get a heavier review). But, since `contentScript` properties can't be verified as safe without manual review, and also aren't at all necessary, they aren't allowed for unlisted add-ons.

magopian: Sorry to be thick, but you mean "they aren't allow for unlisted **non-sideload** add-ons", right?

kmaglione: Yeah, sorry I wasn't clear. They are blanket banned in unlisted, non-sideload add-ons. They are very likely to cause rejection in other add-ons unless they're simple, static strings.

magopian: So, should the message be reworded?

kmaglione: I'm not sure. That is basically what "submitted for automated signing" means, and no-one will actually see this message if it doesn't apply to them. Except people submitting betas, I guess.

magopian: Ah, that's true. And for the moment at least, betas are automatically signed without (automatic) reviews, so it's not even the case.

kmaglione:
I think this takes care of everything. I changed the wording of several messages, and added MDN URLs where they seemed appropriate.

kmaglione:
Fixed failing tests.

magopian:
It's awesome to have all those URLs, thanks! I'm not sure about the shortener though.

r+wc

PR #306, opened 2015-07-21
Don't reprocess messages when pushing/popping package stack.

kmaglione:


magopian:
Is that linked to a bug? What is it about?

kmaglione:
The function that I removed was the root cause of https://bugzil.la/1172148. When I fixed that, I mentioned that this code was unreliable and I planned to get rid of it. I wrote the patch a while ago, and decided to sit on it for a while, until I wrote #305 and only realized by accident that I'd need to update `_merge_messages`. I don't want it to be the cause of further errors.

magopian:
Looks good to me from a code point of view. @mstriemer do you have the time to have a look?

PR #307, opened 2015-07-21
Remove 4360 lines of old compat tests.

kmaglione:


validator/testcases/javascript/instanceproperties.py
```diff
@@ -60,8 +58,8 @@ def _set_HTML_property(function, new_value, traverser):
                 # Everything checks out, but we still want to pass it through
                 # the markup validator. Turn off strict mode so we don't get
                 # warnings about malformed HTML.
-                from validator.testcases.markup.markuptester import \
-                                                                MarkupParser
+                from validator.testcases.markup.markuptester import (
+                    MarkupParser)
```
magopian: thanks

magopian:
I don't understand why/how we don't need this anymore. Stuff that we flagged as obsolete/deprecated in the past is still obsolete and deprecated nowadays, right? Maybe I'm missing something

kmaglione:
These tests are all only useful just before release of a new Firefox version. After that, they'd only be relevant to ancient, cargo-culted code, but anyone using that code should notice the breakage. In any case, these tests are only used for the automated compatibility runs, after which developers don't even see them.

I talked to @jvillalobos  about removing these, and he agreed that they should go.

magopian:
Then r+, I'm all for removing code! Thanks!

PR #309, opened 2015-07-26
Bug 1185021: Part 1 - Convert runtime exceptions into validator errors...

kmaglione:
Sorry for the massive test changes. I went to update the tests for the functions this touched, and they turned out to be a disaster.


validator/errorbundler.py
```diff
@@ -116,6 +120,44 @@ def wrap(self, *args, **kwargs):
...
+        `exc_info` is supplied, the error will be logged.
+
+        If the error is a validation timeout, it is re-raised unless
+        `msg_id` is "validation_timeout"."""
+
+        if exc_info:
+            if (isinstance(exc_info[1], validator.ValidationTimeout) and
+                    msg_id != "validation_timeout"):
+                # These should always propagate to the top-level exception
+                # handler, and be reported only once.
```
magopian: Where is the "be reported only once" part?

validator/errorbundler.py
```diff
@@ -116,6 +120,44 @@ def wrap(self, *args, **kwargs):
...
+                         description or
+                         ("Validation was unable to complete successfully due "
+                          "to an unexpected error.",
+                          "The error has been logged, but please consider "
+                          "filing an issue report here: "
+                          "http://mzl.la/1DG0sFd"),
+                         tier=1)
+
+        # Move the error message to the beginning of the list.
+        assert self.errors[-1] is err
```
magopian: Should we have this here? I'd rather have it in a test.

validator/main.py
```diff
@@ -9,14 +10,15 @@
 def main():
     "Main function. Handles delegation to other functions."
 
+    logging.basicConfig()
```
magopian: Is that needed?

validator/submain.py
```diff
@@ -46,49 +37,50 @@ def prepare_package(err, path, expectation=0, for_appversions=None,
...
-    # Test that the package is an XPI.
-    if package_extension not in (".xpi", ".jar"):
-        if err:
-            err.error(("main",
-                       "prepare_package",
-                       "unrecognized"),
-                      "The package is not of a recognized type.")
-        return False
+    package = None
+    try:
```
magopian: The `try` doesn't need to enclose those following lines. I think it could be moved down to line 60

validator/testcases/content.py
```diff
@@ -297,13 +298,13 @@ def _process_file(err, xpi_package, name, file_data, name_lower,
         try:
             sub_xpi = XPIManager(package, mode="r", name=name,
                                  subpackage=is_subpackage)
-        except Exception:
+        except BadZipfile:
```
magopian: Do we have a good reason to have two different places where we try to open the zipfile, and raise an error if we can't?

validator/testcases/markup/markuptester.py
```diff
@@ -119,18 +116,15 @@ def _feed_parser(self, line):
         line = unicodehelper.decode(line)
 
         try:
-            self.feed(line + "\n")
-        except UnicodeDecodeError, exc_instance:
-            exc_class, val, traceback = sys.exc_info()
             try:
+                self.feed(line + "\n")
```
magopian: We don't need two nested `try/except` for that, do we?

validator/python/copy.py
```diff
@@ -1,433 +0,0 @@
-"""Generic (shallow and deep) copying operations.
```
magopian: Wait, we don't get to have our own custom implementation of `copy`? *sadface*.
Jokes aside, very nice cleanup ;)

tests/test_submain.py
```diff
@@ -1,47 +1,77 @@
...
+from .helper import MockXPI
 
-def test_prepare_package():
-    "Tests that the prepare_package function passes for valid data"
 
-    tp = submain.test_package
-    submain.test_package = lambda w, x, y, z, for_appversions: True
+@mock.patch("validator.submain.test_package")
+def test_prepare_package(test_package):
+    "Tests that the prepare_package function passes for valid data."
```
magopian: `passes for`, is that a typo? I don't understand the sentence

tests/test_submain.py
```diff
@@ -1,47 +1,77 @@
...
+    submain.prepare_package(err, "foo/bar/test.foo")
+
+    assert not test_search.called
+    eq_(len(err.errors), 1)
+    eq_(err.errors[0]["id"], ("main", "prepare_package", "not_found"))
+
+    err = ErrorBundle()
+    submain.prepare_package(err, "foo/bar/test.xml")
+
+    assert not test_search.called
```
magopian: A small comment explaining that it's not being called because the file doesn't even exit?

tests/test_submain.py
```diff
@@ -1,47 +1,77 @@
...
-    assert submain.prepare_package(None, "foo/bar/test.xml") == True
-    submain.test_search = ts
+@mock.patch("validator.submain.test_search")
+def test_prepare_package_extension(test_search):
+    "Tests that bad extensions get outright rejections."
+
+    err = ErrorBundle()
+    submain.prepare_package(err, "foo/bar/test.foo")
+
+    assert not test_search.called
```
magopian: A small comment explaining that it's not being called because it doesn't have a correct extension?

tests/test_submain.py
```diff
@@ -291,7 +315,10 @@ def get_tests(self, tier, type):
 
         self.last_tier = tier
 
-        for x in range(1,10): # Ten times because we care
+        for x in range(1, 10):  # Ten times because we care
+            # ^ Very witty. However, it would be nice to actually know
+            # exactly why we're yielding these ten times.
+
```
magopian: Just to pretend there's many tests? To slow down the test suite? I have no idea, but this is a code smell. I'd try just removing the loop, and see if it breaks anything. If it does, maybe it'll give us a better idea of what this is for?

magopian:
Nice cleanup ;)

r+wc

validator/errorbundler.py
kmaglione: The fact that we re-raise before we actually report it if we're not at the top-level.

validator/errorbundler.py
kmaglione: Fair enough. There's already a test for it.

validator/errorbundler.py
magopian: ah, understood, thanks ;)

validator/main.py
kmaglione: Yeah, otherwise stack traces don't get reported, and we get a `No handlers could be found for logger "amo.validator"` message when running from the command line. (The current release already emits  `No handlers could be found for logger "root"` messages from time to time for the same reason.)

validator/python/copy.py
kmaglione: Heh. It's not even our own implementation. It's just a copy from the standard library. I don't even know why we have it, since I don't think we've ever run a version of Python old enough not to have it.

validator/submain.py
kmaglione: I originally enclosed much less code, yeah, but then I thought that, given the purpose is to catch any internal error in the validator, when at all possible, it made sense to enclose as much code as possible rather than as little (which would be my usual preference).

validator/testcases/content.py
kmaglione: We should probably move this test and reporting to `XPIManager`, but we're not testing the same file twice. In the other instance, we're testing a top-level add-on package, where this is testing a JAR with arbitrary content that belongs to the same add-on.

validator/testcases/markup/markuptester.py
kmaglione: Yeah, we do. This one is just to catch an encoding error and try falling back to ascii if it fails. The outer one is to catch an actual parser error. Arguably, this should just use `unicodehelper.decode`, but I'd rather not make that kind of change in this PR.

tests/test_submain.py
kmaglione: Yeah, I didn't understand it either :( But rewriting all of the docstrings in this file wasn't an adventure I wanted to start down.

tests/test_submain.py
kmaglione: Yeah, fair enough. It took me long enough to figure out that was what it was supposed to be testing...

validator/testcases/markup/markuptester.py
magopian: so you want to catch an error that would happen in this `except` body? Is that what you're saying? If you're using `ignore`, I don't think you can have exceptions. I don't see any other reason to have those nested try blocks. Am I missing something?

kmaglione: This `try-catch` block catches an encoding error on the first try. After that, it force-converts it to unicode (does not try to catch unicode errors) and tries again. Either attempt, though, might raise another parser error, which is what the outer `try-catch` catches.

tests/test_submain.py
kmaglione: I guess what I'm more worried about is that it'll stop breaking something that we need it to break...

PR #310, opened 2015-07-29
Optimize all the things

kmaglione:
Some numbers:

Previous time to complete full run of tests: 14.6s
Current time to complete a full run of tests: 2.9s
Percent improvement: 80%

Previous time to validate https://bugzil.la/1184979: 83s
Current time to validate https://bugzil.la/1184979: 32s
Percent improvement: 61%


kmaglione:
9b51091 doesn't really belong here, but I hit it after trying to time my test add-on after rebasing.

magopian:
That's some really code work here, and very clever of you to tackle the low hanging fruit after having spotted them using a profiler (being curious, which tools did you use? `cProfile` and RunSnakeRun?)

I'm really surprised changing "just those" made such a significant improvement, yay!

r+wc

requirements.txt
```diff
@@ -6,3 +6,4 @@ rdflib==3.2.1
 argparse==1.1
 fastchardet==0.2.0
 mock==1.0.0
+simplejson==3.8.0
```
diox: Curious about this change. app-validator used to use simplejson instead of just json, and we dropped that when we switched to python 2.7 because the performance difference was negligible (and since simplejson was a compiled dependency, it was more annoying to keep it around)

kmaglione:
I used `cProfile` and `cprofilev` for the most part, and `line_profiler` for a few individual functions.

requirements.txt
kmaglione: For the particular add-on I was testing, it saved 5 or 6 seconds overall.

kmaglione:
Cleaned up the, er, cleanup code, and added some tests for it. Fixed other nits.

tests/test_js_spidermonkey.py
```diff
@@ -40,52 +37,84 @@ def test_scripting_snippet():
...
-def test_unicode_escapes():
-    """Tests that unicode/hex escapes are passed through to Spidermonkey."""
+    # Make sure we don't already have a cached shell.
+    JSShell.cleanup()
+    eq_(JSShell.shells, {})
 
-    result = _do_test_raw("var foo = '\\u263a\\u0027\\x27'")
+    process = mock.Mock()
+    process.stdout.readline.return_value = "{}"
+    Popen.side_effect = lambda *args, **kw: process
```
magopian: Why a side effect here, and above a `.return_value = `?

tests/test_js_spidermonkey.py
```diff
@@ -40,52 +37,84 @@ def test_scripting_snippet():
...
+    # Read error. No cached shell.
+    process.stdout.readline.side_effect = IOError
+
+    with assert_raises(IOError):
+        get_tree("hello();", shell=SPIDERMONKEY_INSTALLATION)
+
+    # No new processes created, since we had a cached shell from last time.
+    eq_(Popen.call_count, 1)
+    assert process.stdin.write.called
+    assert process.stdout.readline.called
```
magopian: `assert process.terminate.called` ?

tests/test_js_spidermonkey.py
kmaglione: Because this test assigns a new mock to `process`. I originally had it as a `return_value`, but it wound up too messy.

tests/test_js_spidermonkey.py
```diff
@@ -40,52 +37,84 @@ def test_scripting_snippet():
...
+    process.stdin.write.side_effect = IOError
+
+    with assert_raises(IOError):
+        get_tree("hello();", shell=SPIDERMONKEY_INSTALLATION)
+
+    # Cached shell should have been removed in the last run. New
+    # process should be created.
+    eq_(Popen.call_count, 2)
+    assert process.stdin.write.called
+    assert not process.stdout.readline.called
```
magopian: Also assert here that the process was terminated?

tests/test_js_spidermonkey.py
kmaglione: I originally had it, but the deconstructor doesn't get called immediately in this test, so it fails. There are other tests for the deconstructor, so it didn't seem worthwhile to try to find a way to make that test work here.

validator/decorator.py
```diff
@@ -28,6 +34,29 @@ def wrap(function):
     return wrap
 
 
+def register_cleanup(cleanup):
+    """Register a cleanup function to be called at the end of every validation
+    task. Takes either a callable, or a class with a `cleanup` class method."""
+
+    if not callable(cleanup):
```
magopian: Just a side note: if the class has a `__call__` method, it'll be detected as callable. Shouldn't be an issue for us though, but maybe add a comment?

kmaglione: Yeah, I considered that, but the docstring says `callable`, and I figured that anyone writing a callable class would think of that possibility. I guess it couldn't hurt to be extra specific, though.

magopian:
r+wc

PR #311, opened 2015-07-30
Remove the old regex test implementation

kmaglione:
I've been meaning to do this for a while, for a lot of reasons. It's currently on top of #310, but I decided to keep it separate, since the performance improvement to complexity ratio is much less than any of those changes.

This is the first step in moving the definitions of tests out of files based on the exact type of tests they are (`regex.py`, `instanceactions.py`, `predefinedentities.py`), and into files based on what they're testing (preferences (including pref strings and APIs), HTML generation, compatibility with an app version, ...).


kmaglione:
I'm going to suggest not spending a lot of time reviewing this right now, though. I mainly want the PR here just so it's on the record in case anyone plans to touch that code. I'd be happy with this not landing for another couple of weeks.

magopian:
Ok, please let me know when you want me to review it

kmaglione:
OK, this should be ready now. It now saves about 4s on the benchmark XPI, and lays the groundwork for a lot of cleanup that I have a pending patch for, so it's fairly high priority.

magopian:
needs a rebase

I'm worried that the resulting code is not simpler. It may be shorter (and faster?), but I'm worried that the result is more difficult to grasp, which is an issue.

I understand this is needed to land some other changes that are coming? Could you tell me more about those, maybe on irc?

kmaglione:
The regex test classes are definitely not simple. They sacrifice some simplicity for performance, and the flexibility the performance gives us. I'm hoping to simplify this code more, and remove unneeded features as we get tests better organized. But in particular, what I'm planning to build on this:

1) Move most of the regex tests out of the regex-specific files. For instance, all related security tests live in the same `security.py` file, and import `testcases.regex.javascript` to add the string tests they need. I already have a WIP for this.

2) The `err.notify` function should be used everywhere we accept partial message definitions. If there's a `dangerous` property on a global object definition, it gets passed to `traverser.notify` with a default ID, message, and description.

3) I plan to further simplify `notify`, and just replace `warning`, `error`, and `notice` parameters with `message` and `type` parameters, which is what they get stored as anyway. But that requires fairly broad changes to existing tests, so I didn't want to do it in this commit.

4) I want to split all regexps that apply to JS into a separate test class, and run them directly on .js/.jsm files, as well as things like script nodes and event listener attributes in HTML documents. These already all get passed to the same `scripting.test_js_file` function, so it should simplify many things. The old regex tests were way too slow and monolithic to make this possible.

kmaglione:
Here's a very rough preview of what I'm planning on building on this commit:

https://gist.github.com/fcae0ed2aaa2657a870e

I'm most likely going to replace the `FILE_REGEXPS` array with method call on the `RegexTest` objects, along the lines of `file_regexps.add_test(...)`, and I'm still trying to work through some of the details of the global/interface hook APIs.

But I'm trying to minimize the changes I'm making for now, outside of code moves, so the file is still full of strange things like that `@register_entity` function, and that `call_settimeout` function with one-letter argument names, which directly inspects the parse node...

magopian:
r+wc

EnTeQuAk:

PR #312, opened 2015-07-30
Use python-spidermonkey module for Spidermonkey binary.

kmaglione:


kmaglione:
I think this is ready now. Ideally I'd like to get this merged (but not pushed to production) sometime this week so we can start enabling the validator on Olympia dev builds (especially in Docker) by default.

requirements.txt
```diff
@@ -7,3 +7,4 @@ argparse==1.1
 fastchardet==0.2.0
 mock==1.0.0
 simplejson==3.8.0
+spidermonkey==41.0a2.post1
```
magopian: :+1:  Do you want me to upload that to pyrepo?

tests/test_validate.py
```diff
@@ -16,7 +15,9 @@ def run(self, path, **kwargs):
         self.output = validate(path=path, **kwargs)
 
     def test_metadata(self):
-        """Test that the generated JSON has the appropriate metadata valuees."""
+        """Test that the generated JSON has the appropriate metadata
+        valuees."""
```
magopian: typo: values.

validator/testcases/javascript/jsshell.py
```diff
@@ -64,44 +65,43 @@ class JSShell(object):
...
 
-        return cls.shells[shell]
+        return cls.instance
 
     @classmethod
     def cleanup(cls):
-        """Clears any saved shells, and terminates their Spidermonkey processes
-        if there are no further references."""
-        cls.shells.clear()
+        """Clear our saved instance, and terminates its Spidermonkey process,
```
magopian: `...and terminate its...`

magopian:
If you don't want it pushed to production this week, please only merge after the tag for this week is done (tomorrow).

That's a nice improvement. I'm still a bit confused and lost with all the indirection in the `jsshell` file, but I guess it's needed.

r+wc

requirements.txt
kmaglione: Yeah, we'll need it there. Do you want me to give you a wheel for Linux? PyPI won't accept it.

kmaglione:
If there's anything you want me to explain the rationale for, let me know. It's mostly just a matter of making sure we use a single Spidermonkey process for the entire validation run, but to clean it up at the end, and whenever we run into errors.

requirements.txt
magopian: Done: https://pyrepo.addons.mozilla.org/spidermonkey-41.0a2.post1.tar.gz

Not sure we need the wheel for now. It would be a nice speedup, but if it's a pain to create... travis could do it for us, but if it can't be uploaded to pypi...

PR #314, opened 2015-08-05
Automagically switch to single-quoted strings everywhere.

kmaglione:
This was done using https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/string_fixer.py

It's been widely in use as a pre-commit hook for at least the past six months, uses the python tokenizer to extract strings, and only handles unambiguous cases (i.e., strings without single or double quotes in them, no triple-quoted strings), so it should be very reliable.


magopian:
Seriously?

kmaglione:
Yeah. I keep getting the quote type wrong either in the validator or in olympia because they use different standards. This was done using a script based on the python tokenizer, so the results are reliable. Also, it took me about 15 seconds to do this from start to finish.

magopian:
After discussing with @kmaglione on IRC, he explained that he used the script at https://github.com/pre-commit/pre-commit-hooks/blob/master/pre_commit_hooks/string_fixer.py#20. The line highlighted shows that the script doesn't try to do anything clever if there are enclosed quotes, which is a good thing.

I wouldn't personally have done that change, because it adds noise to the history (think about `git blame`), but why not. @kmaglione is the main dev nowadays, and he thinks this is important and useful.

r+

magopian:
This needs a rebase. I would advise merging it asap, or it'll need rebasing over and over ;)

kmaglione:
Yeah, I'm just trying to decide how to rebase my other PRs on top of it before I pull the trigger.

kmaglione:
Solution:

Modify the `string_fixer.py` script to read from stdin and print to stdout. Save it somewhere in your path as `fix_python_quotes`.

Add the following to `.git/config`:

```ini
[filter "singlequotes"]
	clean = fix_python_quotes
	smudge = fix_python_quotes
```

Add the following to `.git/info/attributes`:

```
*.py	filter=singlequotes
```

Make a dummy commit: `git commit -am Foo.`

Run a rebase, renormalizing all the way, and dropping your dummy commit:

`git rebase -X renormalize single-quotes`

...

Profit?

Right. I think I need a shower.

PR #318, opened 2015-08-09
Metadata version

kmaglione:


kmaglione:
Er, I didn't mean to include that pytest commit...

setup.py
```diff
@@ -1,9 +1,14 @@
 from setuptools import setup, find_packages
 
 
+# Do not import any validator modules here, as that would require a correct
+# dependency environment that setup.py is not guaranteed to have.
+execfile("validator/_version.py")
+
+
```
magopian: I'd rather have something simpler and more widespread like storing the version in the `__init__.py` file, as a `VERSION` or `__version__`.

Here's an example:
https://github.com/django/django/blob/master/setup.py#L32
and https://github.com/django/django/blob/master/django/__init__.py#L3

I wouldn't bother reimplementing the `get_version` though, I think we can manually update both the `VERSION` and `__version__` on each release.

magopian:
r+wc, it's missing a test

setup.py
kmaglione: That was my initial plan. The main problem is that if we do that, we need to make sure `validator` can always be imported without installing any dependencies, since `setup.py` is expected to run without them. And if one of us breaks that, we're not likely to notice, since we'll probably have the dependencies installed anyway. I spent some time looking for alternative ways to single source the version, and everything I read also discouraged against importing `__init__` from `setup.py`. This was the most reasonable alternative I found.

magopian: Well, we have the CI (travis) to tell us if we failed something, right?

kmaglione: To check this, we'd have to add another set of tests that didn't install any dependencies. We do have the PyPI deploy script, which should qualify, but that only runs when we tag, and is broken at the moment anyway.

Also, I'm not even sure it's a good idea for `import validator` to work without installing dependencies.

PR #323, opened 2015-08-27
Make the timeout error high severity

mstriemer:


magopian:
Makes sense, thanks @mstriemer 
r+

mstriemer:
@kmaglione Should all system errors be high severity? https://github.com/mozilla/amo-validator/blob/280dff511976fbafd4799158b7428b41a183bb83/validator/errorbundler.py#L148

PR #324, opened 2015-09-02
Remove newline in eval warning (no bug)

wagnerand:
This changes:

Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the most innocuous of circumstances. Please avoid using

`eval` and the `Function` constructor when at all possible.

into:

Evaluation of strings as code can lead to security vulnerabilities and performance issues, even in the most innocuous of circumstances. Please avoid using `eval` and the `Function` constructor when at all possible.

(effectively just removing the newline in the middle of the sentence)

@kmaglione r?


kmaglione:
r+

PR #326, opened 2015-09-02
Warn about NewTabURL.override (bug 1200822)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1200822

### DIFF
--- a/validator/testcases/javascript/entity_values.py
+++ b/validator/testcases/javascript/entity_values.py
@@ -149,6 +149,28 @@ def nsITransferable_init(traverser):
     return {'return': on_init}
 
 
+@register_entity('NewTabURL.override')
+def NewTabURL_override(traverser):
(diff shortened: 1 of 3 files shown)


validator/testcases/javascript/entity_values.py
```diff
@@ -149,6 +149,27 @@ def on_init(wrapper, arguments, traverser):
     return {'return': on_init}
 
 
+@register_entity('NewTabURL.override')
+def NewTabURL_override(traverser):
+    def on_override(wrapper, arguments, traverser):
+        from validator.testcases.javascript.predefinedentities import (
+            CUSTOMIZATION_API_HELP)
```
magopian: Why the import here and not at the top of the file? If it's to avoid an import loop, please add a comment saying so

magopian:
Looks good to me (minor nit)

PR #328, opened 2015-09-22
Error if sub-XPI is not Mozilla signed (bug 1190704)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1190704

### DIFF
--- a/tests/test_signed_xpi.py
+++ b/tests/test_signed_xpi.py
@@ -59,7 +59,7 @@ def test_other_signed_xpi():
 
 
 def test_mozilla_signed_multi_xpi():
-    """Test that signed packages raise warning"""
+    """Test that mozilla signed inner packages don't raise an error."""
     xpi = MockXPI({
         'META-INF/manifest.mf': 'tests/resources/main/foo.bar',
         'META-INF/mozilla.rsa': 'tests/resources/main/foo.bar',
@@ -69,6 +69,7 @@ def test_mozilla_signed_multi_xpi():
     err = ErrorBundle()
     err.supported_versions = {}
     err.push_state('sub.xpi')
+    err.save_resource('is_multipackage', True)
 
     content.test_signed_xpi(err, xpi)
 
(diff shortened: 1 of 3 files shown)


tests/test_signed_xpi.py
```diff
@@ -56,3 +56,73 @@ def test_other_signed_xpi():
...
+    content.test_signed_xpi(err, xpi)
+
+    assert err.failed()
+    assert not err.warnings
+    assert err.errors
+    eq_(err.errors[0]['id'], ('testcases_content', 'unsigned_sub_xpi'))
+
+
+def test_unsigned_inner_xpi():
+    """Test that inner packages don't need to be signed if not multipackage."""
```
magopian: how can a package be an inner package if it's not in a multipackage?

magopian:
lgtm

tests/test_signed_xpi.py
mstriemer: There's a jar file in test-amo-validator that has `is_nested_package` but isn't a multi package. I'm not quite sure why but maybe because the jar file can't be the root package?

magopian: @kmaglione do you have an idea?

mstriemer: Upon further inspection the add-on that this was failing on is called `chromemanifest-traceback.jar` and it has a file `chrome/inner.jar` so I guess that's what's happening. I suppose the test should have been failing? Would we require the inner `.jar` file to be signed?

mstriemer:
I updated this and test-amo-validator to assume that inner packages must be signed even if they are not multi package extensions (like the one that is a `.jar` with another `.jar` in it from test-amo-validator).

magopian:
I'm not sure about the "sign inner packages that are .jar": I believe we don't want that, and only want to sign `.xpi` files. I think the `.jar` are for themes only (and deprecated?).

Maybe @kmaglione or @jvillalobos will know better

jvillalobos:
The add-on submission page claims to still accept .jar, so I guess we still support it. The file extension should be the only difference, and it's only used for complete themes.

mstriemer:
@magopian Do we currently sign `.jar` files? I'm not sure where that happens.

magopian:
We sign in various places, and indeed I don't think we're filtering on the extension:
1/ we sign when an add-on is reviewed (search for other `sign_file()`): https://github.com/mozilla/olympia/blob/master/apps/editors/helpers.py#L700
2/ we sign automatically if possible: https://github.com/mozilla/olympia/blob/master/apps/devhub/views.py#L1207
But only if the add-on supports an application we want to sign for: https://github.com/mozilla/olympia/blob/master/lib/crypto/packaged.py#L27

mstriemer:
@magopian this seems right then? Inner `.jar` files need to be signed?

magopian:
I guess so ;)
r+

PR #329, opened 2015-09-24
Change trivial unlisted warnings to notices (bug 1172030)

mstriemer:
r? @kmaglione 

This makes the trivial signing severity messages notices instead of warnings. Since there were only these two I opted to do it here rather than in the JS on olympia.


mstriemer:
@kmaglione Any concerns with this approach?

magopian:
r+

@kmaglione if you have any concerns with this approach, please let us know and we'll fixup ;)

PR #331, opened 2015-10-09
Allow unsigned inner jar files (bug 1213163)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1213163

Just want to see the build fail for now.


kmaglione:
We should probably test `.xpi` files regardless of whether it's a multi-package extension, and `.jar` or anything else only if it is.

r+wc

mstriemer:
Could an inner XPI be used in any way if it isn't a multi-package extension?

PR #332, opened 2015-10-19
Add backbone.localStorage and moment.js to libraries

kewisch:
When I run jslibfetcher and build_whitelist, the hashes of various existing libraries are changed. This includes most minimized jquery files, angular version 1.1.0 and 1.1.2 and dojo version 1.5 and 1.6, and some others. I'm not sure if this is a known issue.

Anyway, I reverted the changes to the other hashes and only pushed those I've added.


magopian:
Thanks @kewisch for this ;)

@kmaglione pinging you for the comment above, is that expected?

r+

wagnerand:
I have seen that too. That should not block this PR though.

Not sure if @kmaglione will finish the lib detection/whitelisting PRs, maybe we just skip it and do it in the addons-validator (JS).

PR #334, opened 2015-10-28
Firefox 42 compatibility (pair prog with @cgrebs)

magopian:
Fixes [bug 1214929](https://bugzilla.mozilla.org/show_bug.cgi?id=1214929)
Fixes [bug 1214926](https://bugzilla.mozilla.org/show_bug.cgi?id=1214926)
Fixes [bug 1214921](https://bugzilla.mozilla.org/show_bug.cgi?id=1214921)
Fixes [bug 1214912](https://bugzilla.mozilla.org/show_bug.cgi?id=1214912)

### DIFF
--- a/tests/compat/test_gecko42.py
+++ b/tests/compat/test_gecko42.py
@@ -51,3 +51,36 @@ class TestFX42Compat(CompatTestCase):
             window.mozRequestAnimationFrame();
         """)
         self.assert_compat_error()
+
+    def _test_nsIPermissionManager(self, method):
+        self.run_script_for_compat("""
+            var perms = Cc["@mozilla.org/permissionmanager;1"]
+                          .getService(Ci.nsIPermissionManager);
+            %s
+        """ % method)
+        self.assert_compat_error()
+
+    def _test_ServicesPerms(self, method):
+        self.run_script_for_compat("""
+            Components.utils.import("resource://gre/modules/Services.jsm");
+            var perms = Services.perms;
+            %s
+        """ % method)
+        self.assert_compat_error()
+
+    def test_nsIPermissionManagerMethods(self):
+        methods = ("perms.add(url, 'cookie', 1);",
+                   "perms.addFromPrincipal(url, 'cookie', 1);",
+                   "perms.remove('yahoo.com', 'cookie');",
+                   "perms.removeFromPrincipal('yahoo.com', 'cookie');",
+                   "perms.removeAll();",
+                   "perms.testExactPermission(url, cookie);"
+                   "perms.testExactPermissionFromPrincipal(url, cookie);",
+                   "perms.testPermission(url, cookie);",
+                   "perms.testPermissionFromPrincipal(url, cookie);")
+
+        for method in methods:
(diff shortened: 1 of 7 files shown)


magopian:
Ok so we have issues verifying those modifications: namely in the two following cases:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1214929#c2
- https://bugzilla.mozilla.org/show_bug.cgi?id=1214912#c2

@andymckay, who would be able to give a review (and some ideas on the comments above)?

validator/testcases/javascript/predefinedentities.py
```diff
@@ -9,7 +9,7 @@
 import instanceactions
 from jstypes import JSWrapper
 from validator.compat import FX40_DEFINITION
-from validator.constants import MDN_DOC
+from validator.constants import BUGZILLA_BUG, MDN_DOC
```
mstriemer: `BUGZILLA_BUG` isn't used.

mstriemer:
The second issue should be solved by #335.

magopian:
The second issue is indeed solved by #335, thanks @mstriemer you rock!

magopian:
And we now also fixed the first issue, knowing that it may raise false positives (it'll flag any use of the `parseContentType` methods, whatever they relate to, even if they're not on the `nsINetUtil` service).

tests/compat/test_gecko42.py
```diff
@@ -0,0 +1,86 @@
...
+        """)
+        self.assert_compat_error()
+
+    def _test_nsIPermissionManager(self, method):
+        self.run_script_for_compat("""
+            var perms = Cc["@mozilla.org/permissionmanager;1"]
+                          .getService(Ci.nsIPermissionManager);
+            %s
+        """ % method)
+        self.assert_compat_error()
```
mstriemer: This should probably look like:

```py
self.assert_silent()
self.assert_compat_error()
```

I know that's weird but the validation gets run twice. `self.assert_silent()`, `self.assert_failed()`, etc work against `self.err` and `self.assert_compat_error()` and `self.assert_compat_silent()` work against `self.compat_err`.

validator/testcases/javascript/entity_values.py
```diff
@@ -171,6 +171,33 @@ def on_override(wrapper, arguments, traverser):
...
+@register_entity('nsIPermissionManager.remove')
+@register_entity('nsIPermissionManager.removeFromPrincipal')
+@register_entity('nsIPermissionManager.removeAll')
+@register_entity('nsIPermissionManager.testExactPermission')
+@register_entity('nsIPermissionManager.testExactPermissionFromPrincipal')
+@register_entity('nsIPermissionManager.testPermission')
+@register_entity('nsIPermissionManager.testPermissionFromPrincipal')
+def nsIPermissionManager(traverser):
+    traverser.err.warning(
+        err_id=('testcases_javascript_entity_values',
```
mstriemer: I think you need to set the version you want the compatibility error to happen for in this call.

PR #335, opened 2015-10-28
Force the Cc variable to trigger predefined entities

mstriemer:
This should fix the issue where the test code isn't getting run in #334.


magopian:
looks good to me. It's a shame that we need to use such a hack, but if it works, I guess it's good enough until we have the new shiny addons-validator.

r+

magopian:
Merged the PR because we want that asap, thanks again @mstriemer you made our day!

PR #336, opened 2015-10-29
Don't force xpcom_wildcard if the local_variable isn't a dict

mstriemer:


magopian:
That looks good to me, and does work as expected locally, fixing the [google document viewer addon issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1214921#c3)

r+

PR #337, opened 2015-10-29
Fix compat tests TIERs, newTab.xul test should be a compat error

magopian:
This is a follow-up to PR #334

### DIFF
--- a/tests/compat/test_gecko42.py
+++ b/tests/compat/test_gecko42.py
@@ -38,6 +38,9 @@ class TestFX42Compat(CompatTestCase):
                 'chrome.manifest'))
         content.test_newTab_xul(err, None)
         assert err.failed()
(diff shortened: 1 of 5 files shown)


tests/compat/test_gecko42.py
```diff
@@ -38,6 +38,7 @@ def test_newTab_xul(self):
                 'chrome.manifest'))
         content.test_newTab_xul(err, None)
         assert err.failed()
+        assert err.compat_summary['errors'] == 1
```
mstriemer: I'm probably being overly cautious but can you do something like this instead:

```py
assert not err.errors
eq_(len(err.warnings), 1)
eq_(err.warnings[0]['compatibility_type'], 'error')

mstriemer:
r+wc

PR #341, opened 2015-11-05
Integrate the functional tests from the test-amo-validator repository

magopian:
I don't know why those were in a separate repository, maybe there's a good reason? @mstriemer do you know anything about that?

I've made sure that building the package doesn't include the 25mB of test add-ons.

The `functional_tests` folder is nothing more than a simple copy/paste from the [test-amo-validator repo](https://github.com/mozilla/test-amo-validator).


README.md
```diff
@@ -321,7 +321,23 @@ The same rule applies for the end of a file and for files with only one line.
...
+```
+
+Then make a cup of tea while all of those tests run. It takes a while. If you
+have more than two cores on your machine or you don't mind pwnage, you can try
+to increase the number of parallel processes used for testing.
+
+If you just want to run all the tests at once, run
+
+```bash
+nosetests --processes=2
```
mstriemer: So this runs functional tests by default, then? If they indeed take longer which I think they do I'm not super keen on that. Is there some way to tell `nosetests` to ignore the functional tests by default?

magopian:
As discussed with @mstriemer on IRC, there's a clear drawback of having those functional tests in a separate repo: it's much more difficult to keep it in sync with the amo-validator repo (eg you need to fix a functional test for a amo-validator PR, you can't do that unless you push the fix to the test-amo-validator repo, on master, and thus potentially break any other amo-validator PRs).

We haven't found any drawbacks to merging this (apart from not being able to run those functional tests on any random version of the amo-validator, but I don't believe we have this use case).

I'm thus merging this now.

README.md
magopian: ah, merged this PR before seeing this comment. Yes, by default it'll run all the tests. I'll try to address that in a follow up PR.

mattbasta:
Hey 

I originally built the validator (sorry). The reason why the two repos existed was because (IIRC) there were far more tests long long ago, and there were weird technical limitations with the Jenkins server we had at the time. Kids these days and their TravisCI... 

But there were a lot more tests, I think. At some point they started ripping out features from Firefox and the like, and so many of the tests became irrelevant and were pruned.

PR #343, opened 2015-11-09
Fixed path to appversions.json

wagnerand:
...when the executable is called from another directory.

@kumar303 r?

### DIFF
--- a/validator/validate.py
+++ b/validator/validate.py
@@ -12,8 +12,7 @@ import submain
 
 
 def validate(path, format='json',
-             approved_applications=os.path.join(os.path.dirname(__file__),
-                                                'app_versions.json'),
+             approved_applications=None,
              determined=True,
              listed=True,
              expectation=PACKAGE_ANY,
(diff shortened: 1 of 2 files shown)


validator/validate.py
```diff
@@ -58,6 +57,10 @@ def validate(path, format='json',
                          overrides=overrides, for_appversions=for_appversions)
     bundle.save_resource('is_compat_test', compat_test)
 
+    if (approved_applications is None):
```
mstriemer: No need for the parens here.

validator/validate.py
```diff
@@ -58,6 +57,10 @@ def validate(path, format='json',
...
+        approved_applications = os.path.join(os.path.dirname(__file__),
```
mstriemer: This looks like it doesn't pass PEP8 but I don't see a build failure. Maybe we aren't running PEP8 tests...

This should be:
```py
approved_applications = os.path.join(
    os.path.dirname(__file__),
    'app_versions.json')
# or
approved_applications = os.path.join(os.path.dirname(__file__),
                                     'app_versions.json')

mstriemer:
r+wc

wagnerand:
All done, thanks.

PR #345, opened 2015-11-09
Update app versions

wagnerand:
Updated according to https://addons.mozilla.org/en-US/firefox/pages/appversions/ (excl. Nightlies)

r?


mstriemer:
Thanks

PR #346, opened 2015-11-09
Updated bootstrap.js whitelist

wagnerand:
This updates the `addon-sdk` submodule to tag `1.17` which is needed to generate `validator/testcases/jetpack_data.txt` that contains recent versions of `bootstrap.js`

`tests/resources/bootstrap.js` is identical to https://github.com/mozilla/addon-sdk/blob/1.17/app-extension/bootstrap.js


validator/testcases/jetpack_data.txt
```diff
@@ -35,7 +35,13 @@
 ./app-extension/bootstrap.js firefox-24-06-06-2013 314b849906e2d9d62cd942f90281f2c4d5fe4385b377ba7b2578185ef9f8d58c
 ./app-extension/bootstrap.js firefox-24-06-12-2013 314b849906e2d9d62cd942f90281f2c4d5fe4385b377ba7b2578185ef9f8d58c
 ./app-extension/bootstrap.js firefox-24-dev 314b849906e2d9d62cd942f90281f2c4d5fe4385b377ba7b2578185ef9f8d58c
-./data/bootstrap.js 0.0.12 63d6b92dc63c78fdb745ec4b2bdf5bc7926b3de90a5058d64b6dbcd7eb72dd26
```
mstriemer: It seems strange that this went away but nothing else did. Did that version never actually exist?

tests/test_jetpack.py
```diff
@@ -70,7 +70,7 @@ def test_bad_harnessoptions():
 def test_pass_jetpack():
     """Test that a minimalistic Jetpack setup will pass."""
 
-    harnessoptions = {'sdkVersion': '1.8-dev',
+    harnessoptions = {'sdkVersion': '1.17',
```
mstriemer: I have no idea what is happening in this file.

mstriemer:
What is the expected end result here? I checked this out and reverted the changes to `validator/testcases/jetpack_data.txt` and `validator/testcases/jetpack_data.txt.pickle`and the tests still pass.

validator/testcases/jetpack_data.txt
wagnerand: I am not sure why that version was in there, there never was a v0.0.12.

tests/test_jetpack.py
wagnerand: The tests use `tests/resources/bootstrap.js`, which is jetpack 1.17 now (was 1.8-dev before), therefore all those need to be updated, otherwise the tests fail.

wagnerand:
I was only able to produce `validator/testcases/jetpack_data.txt.pickle` after upgrading to a newer version of jetpack, which I what the first commit does.

mstriemer:
Oh, I see. Sounds good. r+

wagnerand:
Thank you!

PR #347, opened 2015-11-09
Update to jetpack 1.17

wagnerand:
This is needed to generate a `validator/testcases/jetpack_data.txt` that contains recent versions. 

Additionally, the submodule used `1.8-dev`, which caused several test failures when trying to generate the `jetpack_data.txt`. It has been updated to `1.17`.

`tests/resources/bootstrap.js` is identical to https://github.com/mozilla/addon-sdk/blob/1.17/app-extension/bootstrap.js

r?


wagnerand:
This is included in #346 that depends on it.

PR #354, opened 2015-11-13
Flag usage of XPIProvider and AddonManagerInternal, bug 1200929

EnTeQuAk:


validator/testcases/regex.py
```diff
@@ -262,6 +262,27 @@ def js_tests(self):
                 flags=re.I)
 
 
+@register_generator
+class XPIProviderRegexTests(RegexTestGenerator):
+    """
+    These regex tests ensure that developers are not using the XPIProvider
+    or AddonManagerInternal symbols
```
magopian: Sentences should end with a `.` (same below)

tests/test_regex.py
```diff
@@ -50,6 +50,18 @@ def test_dom_mutation_fail():
...
+def test_bug_1200929():
+    """Test that XPIProvider and AddonManagerInternal are not used."""
+
+    real_world_usage = (
+        'XPIProviderBP = Cu.import("resource://gre/modules/XPIProvider.jsm", {});')
+
+    assert _do_test_raw(real_world_usage).failed();
+    assert _do_test_raw('// XPIProvider').failed();
+    assert _do_test_raw('AddonManagerInternal.getAddonByID(1234);').failed();
+    assert _do_test_raw('// AddonManagerInternal').failed();
```
magopian: Is there a way to test the signing severity here?

magopian:
r+wc

muffinresearch:
@EnTeQuAk If you work on adding new functionality to the amo-validator it's always useful to add an issue to addons-validator pointing at related bugs etc. That way it helps us to ensure we don't miss any incoming new features because we're working from http://mozilla.github.io/addons-validator/

EnTeQuAk:
I will do that once I have feedback in the bugzilla bug and once the actual implementation is okay.

EnTeQuAk:
@magopian care for a 2nd review? I updated the error message and added signing severity tests.

tests/test_regex.py
```diff
@@ -50,6 +50,30 @@ def test_dom_mutation_fail():
...
+    """Test that XPIProvider and AddonManagerInternal are not used."""
+
+    real_world_usage = (
+        'XPIProviderBP = Cu.import("resource://gre/modules/XPIProvider.jsm", {});')
+
+    err = _do_test_raw(real_world_usage)
+    assert err.failed()
+    assert err.signing_summary['high'] == 1
+
+    err = _do_test_raw('// XPIProvider')
```
magopian: So this is supposed to fail because it's a regex test, and it can't detect if it's in a comment, right? If so, I'd simply drop this test, because it's not something we _want_, it's something we have, and we don't care about.

tests/test_regex.py
```diff
@@ -50,6 +50,30 @@ def test_dom_mutation_fail():
...
+    assert err.failed()
+    assert err.signing_summary['high'] == 1
+
+    err = _do_test_raw('AddonManagerInternal.getAddonByID(1234);')
+    assert err.failed()
+    assert err.signing_summary['high'] == 1
+
+    err = _do_test_raw('// AddonManagerInternal')
```
magopian: Same here. If somebody gets back to this code in the future and sees, they might wonder why we wanted this specifically ;)

magopian:
r+wc, thanks!

EnTeQuAk:
Squashed and merged in https://github.com/mozilla/amo-validator/commit/1e8bad7e892c4cda8de4e81c58cb0c32e55a9356

PR #356, opened 2015-11-18
bump spidermonkey

magopian:
We need a version greater than 43.0a2 for [bug 1224444](https://bugzilla.mozilla.org/show_bug.cgi?id=1224444)

### DIFF
--- a/tests/test_js_syntax.py
+++ b/tests/test_js_syntax.py
@@ -120,6 +120,44 @@ def test_blocks_evaluated():
         yield test, block
 
 
+def test_generators():
+    """
+    Tests that generators inside simple `function`s (opposed to `function*`)
+    are not raising syntax errors.
+    We use `version(180)` in `testcases.javascript.jsshell.JSShell` so we don't
+    use the latest "version" (185 at the time of this writing) when running
+    spidermonkey. See
+    https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/JSAPI_reference/JSVersion)  # noqa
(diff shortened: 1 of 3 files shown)


tests/test_js_syntax.py
```diff
@@ -69,10 +69,10 @@ def test_blocks_evaluated():
     EVIL = 'eval(evilStuff)'
     BLOCKS = (
         'function foo() { %s; }',
-        'function foo() { %s; yield 1; }',
-        'function foo() { yield %s; }',
+        'function* foo() { %s; yield 1; }',
+        'function* foo() { yield %s; }',
         'var foo = function () { %s; }',
-        'var foo = function () { %s; yield 1; }',
+        'var foo = function* () { %s; yield 1; }',
```
magopian: Those changes are needed for the tests to pass with this version of spidermonkey. Otherwise they raise a syntax error instead.

According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/Legacy_generator_function, "The legacy generator function is a SpiderMonkey-specific feature, and will be removed at some point." (thanks @muffinresearch for digging that up).

magopian:
@kmaglione would you mind giving this a quick r? please? I'm worried about the syntax change, should we simply go with it?

magopian:
As a stop-gap, i've added a `version(180)` to force the version of js (to not be 185, which deprecates the `function` for generators, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/Legacy_generator_function)

EnTeQuAk:
r+

PR #357, opened 2015-11-20
Revert "Whitelist empty file (bug 1224812)"

wagnerand:
This needs to be backed out as the validator not only iterates through files but also folders to find libraries (ugh).
Fixing that issue alone is not sufficient, because we probably do want the validator to recognize empty files, but they should not be part of the (html) report. They should just be greyed out in the source viewer.

r?


magopian:
r+

wagnerand:
Thanks.

PR #360, opened 2015-11-30
Add signing warning if an addon registers to 'newtab-url-changed'.

EnTeQuAk:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1216574

### DIFF
--- a/tests/test_js_signing.py
+++ b/tests/test_js_signing.py
@@ -495,3 +495,33 @@ class TestSearchService(TestCase, RegexTestCase):
             tab.attach({ contentScript: 'doc.createElement("script")' });
         """)
         self.assert_failed(with_warnings=[warning])
+
+
+class TestObserverEvents(TestCase, RegexTestCase):
+    """Tests that code related to observer events trigger warnings."""
+
+    def test_on_newtab_changed(self):
+        """
+        Tests that adding an observer to the `newtab-url-changed` event triggers
+        a high signing severity warning.
+        """
+        patterns = (
+            'Services.obs.addObserver(NewTabObserver, "newtab-url-changed", false);',
+            (
+              'var foo = Cc["foo"].getService(Ci.nsIObserverService);'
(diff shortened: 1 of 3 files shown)


validator/testcases/javascript/entity_values.py
```diff
@@ -171,6 +171,38 @@ def on_override(wrapper, arguments, traverser):
     return {'return': on_override}
 
 
+@register_entity('nsIObserverService.addObserver')
+def nsIObserverService_addObserver(traverser):
+    def on_addObserver(wrapper, arguments, traverser):
+        if not arguments:
+            return
```
magopian: what's that for?

tests/test_js_signing.py
```diff
@@ -495,3 +495,33 @@ def test_contentScript_static_values(self):
...
+    def test_on_newtab_changed(self):
+        """
+        Tests that adding an observer to the `newtab-url-changed` event triggers
+        a high signing severity warning.
+        """
+        patterns = (
+            'Services.obs.addObserver(NewTabObserver, "newtab-url-changed", false);',
+            (
+              'var foo = Cc["foo"].getService(Ci.nsIObserverService);'
+              'foo.addObserver(NewTabObserver, "newtab-url-changed", false);'
```
magopian: which line is the one being caught? It seems you're using two different ways to add the observer, but in the same script, don't you want to have one test for each?

magopian: oh, wow, sorry, i'm just blind

magopian: I know it would add useless and strange indentation if it was "for real", but I think I'd rather have a multiline string here (and no parens). It would make it more readable ;)

magopian:
r+wc

validator/testcases/javascript/entity_values.py
EnTeQuAk: A shortcut to avoid edge-cases below if there are no arguments and stuff.

PR #362, opened 2015-12-04
Flag all instances of mozFetchAsStream for Firefox 43

EnTeQuAk:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1226306


magopian:
Looks good to me. @mstriemer would you mind giving this a quick look?

EnTeQuAk:
@mstriemer while figuring out what this app_versions thing is, I found that it might be appropriate to do something like https://gist.github.com/EnTeQuAk/ce68acc558edd6db158f but I don't know about real world implications :confused:

mstriemer:
Looks good. r+

As far as changing the way the app versions are set I wouldn't worry about it. I normally add 5 of them at a time which takes a few minutes and is good for most of a year. I agree that it is strange but long live addons-validator!

EnTeQuAk:
hehe, alrighty. Thanks @mstriemer :smiley:

PR #363, opened 2015-12-04
Remove old jetpack and jpm files from allow list

wagnerand:
This removes all known Add-on SDK files older than 1.16 from the allow list.
Also it removes all jpm bootstrap.js files older than jpm-core 0.11 as that version fixes the var/let scope change.

The rest are fixes to test failures that were caused by tests relying on very old Add-on SDK allow list entries.

r?


tests/test_content.py
```diff
@@ -327,7 +327,7 @@ def test_subpackage_metadata_preserved():
     err2 = ErrorBundle()
     err2.detected_type = PACKAGE_EXTENSION
 
-    submain.test_package(err1, xpi1, 'jetpack-1.8-outdated.xpi')
+    submain.test_package(err1, xpi1, 'jetpack-1.16-outdated.xpi')
```
magopian: Given the name of the XPI, wasn't it there in the tests on purpose to test outdated XPIs?

wagnerand: It was. 1.17 is the current version, that makes 1.16 an outdated one :)
I cannot use a version older than 1.16 as the tests rely on the hashes
being in the allow list.

But for various reasons, we don't allow 1.15 or older, that's why I removed
all those hashes from the allow list.

magopian: Ok, fair enough ;)

magopian:
r+, thanks!

wagnerand:
Thanks for reviewing!

PR #372, opened 2015-12-29
Flag 'xpinstall.signatures.required'

wagnerand:
r?


mstriemer:
It looks like this will add a high-severity warning which means that the add-on will still be auto-signed if it is unlisted. Should this be an error instead?

mstriemer:
I guess you could get around this check so it's fine as a warning.

wagnerand:
I agree. The pref will be gone soon anyway, but in the meantime, we need to raise awareness for reviewers during the review.

mstriemer:
r+

wagnerand:
Thanks!

PR #373, opened 2016-01-12
Add compat error for getAllStyleSheets, fixes #1164

EnTeQuAk:


magopian:
r+

tests/compat/test_gecko44.py
```diff
@@ -0,0 +1,15 @@
...
+    """Test that compatibility tests for Gecko 44 are properly executed."""
+
+    VERSION = FX44_DEFINITION
+
+    def test_mozFetchAsStream(self):
+        self.run_script_for_compat(
+            'let stylesheets = window.getAllStyleSheets(config.browser.contentWindow);'
+        )
+        self.assert_silent()
+        self.assert_compat_error()
```
muffinresearch: How does this know that it's the specific compat error you're looking for?

EnTeQuAk: There isn't a specific error id so we can't use that. It's mostly the fact that `run_script_for_compat` will only run the code against the fx44 version and only this code so the chance that there's another error is relatively thin. tldr; it doesn't, I can improve it a tiny bit though and check the message too.

We simply don't have that anywhere else though.

EnTeQuAk: I'm verifying this manually currently.

EnTeQuAk:
merged in master

PR #380, opened 2016-01-22
Set appversion for Firefox 42 compat tests

mstriemer:
More commits to come so we see the test pass/fail progression.


kumar303:
r+ from me so far (pending additional tests).

My validator brain is a little dusty but nothing popped out at me.

PR #381, opened 2016-01-25
Fail for add-ons built with cfx

wagnerand:
r? @mstriemer


mstriemer:
I briefly skimmed this and it looks reasonable but I really didn't get a good look.

Can you add a functional test that a real XPI from cfx fails?

wagnerand:
Test added.

mstriemer:
r+

I'm hoping this has been communicated somehow through compatibility or blog posts or something? Seems like something that could catch a lot of people by surprise.

wagnerand:
I suggested writing a blog post. For now, we have https://developer.mozilla.org/en-US/Firefox/Releases/44 and https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/cfx

PR #386, opened 2016-01-27
Catch target application errors for WebExtension (min version, max version, applications)

### LINKED ISSUE #385: Catch "invalid max version number" for WebExtensions
This is following https://github.com/mozilla/olympia/issues/1338

It seems an invalid max version for an application in a WebExtension's `manifest.json` isn't caught, and results in a weird unrelated error on olympia.

We should catch such an invalid max version just as we do it for a legacy extension.

magopian:
Fixes #385


tests/test_json_parser.py
```diff
@@ -0,0 +1,53 @@
...
+manifest_json = """{
+        "name": "My Awesome Addon",
+        "version": "1.25",
+
+        "applications": {
+            "gecko": {
+                "id": "my@awesome.addon"
+            }
+        }
+    }"""
```
EnTeQuAk: Can we put them into actual json files?

tests/test_rdf.py
```diff
@@ -41,16 +40,6 @@ def test_namespacing():
     assert str(r.uri('bar', 'abc')) == 'abc#bar'
 
 
-def test_namespacing():
```
EnTeQuAk: Wait, why is this test gone again?

EnTeQuAk: `test_get_applications` uses the same rdf file but it doesn't validate the uri or namespace or am I missing something?

validator/json_parser.py
```diff
@@ -0,0 +1,23 @@
+import json
+
+from validator.constants import FIREFOX_GUID
+
+
+class ManifestJsonParser():
```
EnTeQuAk: you forgot `(object)`

EnTeQuAk:
r+wc from what I understand, I like the cleanups along the way

tests/test_json_parser.py
magopian: What for?

tests/test_rdf.py
magopian: `test_namespacing` was actually present twice in the same file (check line 33)

validator/json_parser.py
magopian: What for?

tests/test_json_parser.py
EnTeQuAk: On second thought, you're right. We have json strings all over the place. It just felt better to not put large amounts of JSON into the module for sake of clarity.

tests/test_rdf.py
EnTeQuAk: my bad, no worries then

validator/json_parser.py
EnTeQuAk: Because not inheriting from `object` creates an old-style class in Python 2 which, among a few other things causes `type` to give different results:

```python
>>> class Foo: pass
... 
>>> type(Foo())
<type 'instance'>
```

Compared to new-style classes:

```python
>>> class Bar(object): pass
... 
>>> type(Bar())
<class '__main__.Bar'>
```

Also they behave differently in multiple-inheritance scenarios. You should always prefer new-style classes which are the default in Python 3.

EnTeQuAk: Also see https://docs.python.org/release/2.5.2/ref/node33.html for more information.

magopian: ah, I remember now, `object` is still needed for new style classes in py2 :)

PR #393, opened 2016-02-16
Upgrade cssutils to 1.0.1 to fix unicode space parsing.

EnTeQuAk:
Fixes bug 1244644


diox:
r+ but we'll need that upgrade in addons-server as well.

EnTeQuAk:
Yeah, I'll upgrade it there together with `amo-validator` package in the next step. Thanks for the review!

PR #394, opened 2016-02-16
Make min/max versions in Install Manifests (install.rdf) mandatory

diox:
This seem to be what was always intended: addons-server doesn't like missing min/max versions, MDN says they are mandatory ("minVersion, and maxVersion are all required").

The test was ambiguous because `_do_test()` helper only checks if `failed()` returned True, but it by default simple warnings will cause that, not just errors.

Fixes mozilla/addons-server#1390

### DIFF
--- a/tests/test_targetapplication.py
+++ b/tests/test_targetapplication.py
@@ -1,7 +1,5 @@
-import json
 from mock import patch
 import validator.testcases.targetapplication as targetapp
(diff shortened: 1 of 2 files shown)


diox:
Note: this does not affect web extensions, which can still omit min/max versions. Only XPIs with `install.rdf` should be affected by this change.

EnTeQuAk:
Looks good, r+

PR #400, opened 2016-03-01
Add maxversion * for android as well

wagnerand:
This is to ensure default jpm based add-ons with maxversion "*" would be accepted.

@kumar303 r?


wagnerand:
Ok, that does not work as expected due to a stupid test :-S

(test_validate.py#test_app_versions tests a dummy app_versions.json file tests/resources/test_app_versions.json which of course does not have the real appversions and therefore we cannot just add "*" to entry no 7)

Can we instead just add "*" to `validator/app_versions.json`?

kumar303:
> Can we instead just add "*" to validator/app_versions.json?

Yes, I don't see why not. Since Firefox is no longer strict about it then it should be fine.

validator/constants.py
```diff
@@ -43,9 +43,9 @@
 with open(os.path.join(os.path.dirname(__file__), 'app_versions.json')) as avs:
     APPROVED_APPLICATIONS = json.load(avs)
 
-# Add the '*' fake max version. Used specifically for WebExtensions, see
-# https://github.com/mozilla/amo-validator/issues/390.
+# Add the '*' fake max version for Desktop and Android.
 APPROVED_APPLICATIONS['1']['versions'].append('*')
+APPROVED_APPLICATIONS['7']['versions'].append('*')
```
kumar303: I know you are just copying the last person but this was a terrible approach IMO. Instead, what if you did it like this:
````python
for app_id, config in APPROVED_APPLICATIONS.items():
    if config['name'] in ('Firefox', 'Android'):
        APPROVED_APPLICATIONS[app_id]['versions'].append('*')
````

File: validator/validate.py [hunk 2]
```diff
@@ -66,9 +66,9 @@ def validate(path, format='json',
         # path (string).
         with open(approved_applications) as approved_apps:
             apps = json.load(approved_apps)
-        # Add the '*' fake max version. Used specifically for WebExtensions,
-        # see https://github.com/mozilla/amo-validator/issues/390.
+        # Add the '*' fake max version for Desktop and Android.
         apps['1']['versions'].append('*')
+        apps['7']['versions'].append('*')
```
kumar303: this code shouldn't be copy/pasted. Maybe it's only needed in `validate()` and not needed in `validateor/constants.py` ? If it's really truly needed in both places you could make a shared utility function to add the `*`.

kumar303:
r-

I have suggested a better approach

validator/constants.py
wagnerand: This is going to go away if we just add "*" to app_versions.json, hence my initial question. :)

validator/validate.py
wagnerand: This will also be going away.

validator/constants.py
kumar303: Oh, I thought you originally meant to patch the real [app_versions.json](https://github.com/mozilla/addons-server/blob/master/src/olympia/devhub/tasks.py#L386) we use on the server. In that case, you'd have to patch it in both places.

wagnerand: You mean https://addons.mozilla.org/en-US/firefox/pages/appversions/ ? It already has "*" for Desktop, we can easily add it for Android as well in the admin tools.

wagnerand:
I updated the patch. I didn't completely get https://github.com/mozilla/amo-validator/pull/400#discussion_r54652682 Are you saying I need to patch something in addons-server code as well? (besides adding "*" to the valid Android appversions via the admin panel).

validator/constants.py
kumar303: Yes, I meant that page. Okay, if you add a star for Android through the admin tools then that should fix it.

kumar303:
Looks good to me.

r+

kumar303:
regarding https://github.com/mozilla/amo-validator/pull/400#discussion_r54652682 , yes, adding star to Android through the admin tools will fix it.

PR #402, opened 2016-03-04
Make the cfx test a waning for compat checks

mstriemer:
Fixes mozilla/addons-server#1847.

I _think_ this should fix the issue. But I haven't tested it in olympia because I really don't know how and I don't have the right data.


mstriemer:
Closing in favour of #403.

PR #406, opened 2016-03-10
Update links to point to github

wagnerand:
Note: This makes us lose the template unless we use https://github.com/blog/2111-issue-and-pull-request-templates but that would make it the default for all newly filed issues.

r?

### DIFF
--- a/validator/errorbundler.py
+++ b/validator/errorbundler.py
@@ -152,7 +152,7 @@ class ErrorBundle(object):
                     'to an unexpected error.',
                     'The error has been logged, but please consider '
                     'filing an issue report here: '
-                    'http://mzl.la/1DG0sFd'),
+                    'https://bit.ly/1paBbSy'),
                    tier=1, **kw)
 
         # Move the error message to the beginning of the list.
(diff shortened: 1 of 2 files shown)


muffinresearch:
I wonder if we should consider directing user-bug reports to mozilla/addons instead of mozilla/addons-server. Just because we're going to get more repos in the not too distant so a bug for AMO will not necessarily mean it's addons-server. 

We could add the templates there instead. This would mean that user bugs would need to be referenced from the repo where the fix is being made but cross-project referencing for github issues does work well so that shouldn't be an issue.

tofumatt:
That's an interesting notion. It means addons becomes a place of bug triage and we might need to re-file bugs (though because we have edit permissions we can easily just copy a user's original markdown and move the contents to a new bug easily--or just reference the original bug).

I'm cool with that idea. Savvy users can still file bugs in components as they please, but it prevents novices from getting bogged down in minutia when they just want to report something. :+1:

muffinresearch:
Yeah I think referencing is likely to be nicer than having to copy & paste.

muffinresearch:
@wagnerand OK let's go for pointing to mozilla/addons for issues. I'll sort out something for the readme there.

wagnerand:
Done.

PR #315, opened 2015-08-05
Fx 40 compat

mstriemer:
r? @kmaglione @magopian

PR #412, opened 2016-04-08
Add Firefox 47 compatibility checks

EnTeQuAk:
r? @diox @mstriemer


validator/testcases/regex.py
```diff
@@ -558,6 +558,67 @@ def tests(self):
...
+                'The gDevTools.jsm module should no longer be used. '
+                'The object should now be loaded using '
+                'require("devtools/client/framework/devtools"). See '
+                '%s for more information.' % BUGZILLA_BUG % 1188405
+            ),
+            log_function=self.err.warning,
+            compat_type='warning')
+
+        yield self.get_test(
+            r'\b(fuelIApplication|Application)\b',
```
diox: Just `Application` ? That seems overzealous.

validator/testcases/regex.py
```diff
@@ -558,6 +558,67 @@ def tests(self):
...
+            ),
+            log_function=self.err.warning,
+            compat_type='error')
+
+        yield self.get_test(
+            r'\b(listTokens|listSlots)\b',
+            'listTokens() and listSlots() now return nsISimpleEnumerator instead of nsIEnumerator.',
+            (
+                'listTokens() and listSlots() now return nsISimpleEnumerator instead '
+                'of nsIEnumerator. See %s for more information.' % BUGZILLA_BUG % 1220237
```
diox: I've read the bug and I'm not sure I understand. Wouldn't this flag add-ons using those methods in a way that still works ?

validator/testcases/regex.py
EnTeQuAk: yeah, I asked that in the ticket too to be sure - so I might change that once I get an answer.

validator/testcases/regex.py
EnTeQuAk: This flags addons as a compatibility-error so that they cannot be upgraded to fx47 if they don't change it. Is it that what you mean?

diox: The way I understand the gecko change, it's possible some add-ons do not need to change anything, if their code was already compatible.

EnTeQuAk: Right but I mostly implemented along https://github.com/mozilla/addons-server/issues/2228 which says "compatibility error" and that's what we do here. I suppose they need to be checked for sure but it could be that there are some false positives.

Checking wether or not the code already uses `listTokens` / `listSlots` correctly is very complicated and might not even work imho.

diox: You mean mozilla/addons-server#2220 :)

Yeah, your code looks right - I'd like to confirm with @jvillalobos that the false positives are not going to be an issue just to be sure.

tests/compat/test_gecko47.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self.run_script_for_compat(script)
+
+        assert not self.compat_err.notices
+        assert not self.compat_err.errors
+
+        # There are other checks because using nsIX509CertDB is
+        # potentially dangerous.
+        assert len(self.compat_err.warnings) == 4
+        assert self.compat_err.warnings[3]['compatibility_type'] == 'error'
```
mstriemer: It might be safer to find this error by its id rather than take the fourth one.

validator/testcases/regex.py
mstriemer: Looking in mxr [1] they all seem valid to me. But I haven't looked into each. Looks like there are only a handful of add-ons affected by this.

[1] https://mxr.mozilla.org/addons/search?string=listSlots%28%29&find=&findi=&filter=%5E%5B%5E%5C0%5D*%24&hitlimit=&tree=addons

mstriemer: Some of them even list the type interestingly enough:

```
 var /* nsIEnumerator */ slots = module.listSlots();

File: tests/compat/test_gecko47.py [hunk 4]
```diff
@@ -0,0 +1,183 @@
...
+
+        assert len(self.compat_err.warnings) == 1
+        assert self.compat_err.warnings[0]['compatibility_type'] == 'error'
+        assert self.compat_err.warnings[0]['message'].startswith(message)
+
+    def test_gDevTools_moved(self):
+        """https://github.com/mozilla/addons-server/issues/2219"""
+        message = 'The gDevTools.jsm module should no longer be used.'
+
+        script = 'Components.utils.import("resource:///modules/devtools/client/framework/gDevTools.jsm", ainspectorSidebar);'
```
mstriemer: This looks like slightly more than 80 characters ;)

tests/compat/test_gecko47.py
EnTeQuAk: yeah, I'll change this.

tests/compat/test_gecko47.py
mstriemer: Looks like there's a bunch over 80. This codebase isn't exactly PEP8 compliant but we should still try to do so with new code IMO. The stuff that's about 100 lines is probably fine but this one is particularly long.

validator/testcases/regex.py
mstriemer: It does seem like it could have a lot of false positives. I added the deprecation validations in https://github.com/mozilla/amo-validator/commit/beadb38eb7ed404b4966e085e45fa5d2ac24b11c, they should probably be removed now. I was also checking `extIApplication` which according to https://github.com/mozilla/amo-validator/pull/315 was also required. I'm assuming it was removed as well.

mstriemer:
I was a little surprised that there were multiple tests in each function, I probably would have named each test specifically. In the end I'm not sure the individual tests would help and this keeps the code cleaner with one test mapping to one validation so I'm cool with it.

It's really unfortunate we need to use regex tests for all of this stuff but it's really the only way to catch the majority of it.

r+wc

validator/testcases/regex.py
EnTeQuAk: Right, I'm waiting for an answer in https://github.com/mozilla/addons-server/issues/2218 to clarify wether or not this might have been simply a typo and we should flag `extIApplication` instead of `Application` which to me would make sense since we flagged that on deprecation.

validator/testcases/regex.py
EnTeQuAk: @jvillalobos mind answering wether or not possible false positives are a concern regarding `listSlots` and `listTokens`?

tests/compat/test_gecko47.py
EnTeQuAk: you're so right, I kinda noticed that too when I reviewed the PR just now. I know why I love tools like pytest-pep8 and others :laughing:

EnTeQuAk:
Yeah, some of those flags could probably also be implemented with proper ast walking and stuff but I did that with magopian in the past a few times and it was horrible and hard to match everything we actually wanted so I opted for simple regexes here.

validator/testcases/regex.py
EnTeQuAk: We're flagging `extIApplication` now too.

PR #413, opened 2016-04-13
Remove old angular versions from approved-list

wagnerand:
For background, see https://github.com/mozilla/dispensary/issues/122

r? @mstriemer


mstriemer:
r+

mstriemer:
I added this as a blocker to the push tomorrow.

wagnerand:
Thanks, but not all parts are set in place yet. I'd rather have this go out with the #414. I am ok with this landing on prod next week. I am not sure if you are ok with cherry-picking #414 as well for this week. Either is fine.

andymckay:
If it is being cherry picked, we'll need a package made and the version in addons-server bumped.

mstriemer:
The warnings in #414 seems like a good idea, I'd agree it's too big to cherry-pick. I'd say it's a good idea to at least not accept old angular versions since this is pretty small and could catch some stuff.

PR #414, opened 2016-04-13
Add warning list for third party libraries

wagnerand:
r? @mstriemer


wagnerand:
Note: Some hashes apparently changed, other hashes are in fact hashed 404 pages. This will be cleaned up in a follow-up patch. For now, they don't make anything worse.

extras/update_hashes.py
```diff
@@ -0,0 +1,15 @@
+import glob
+import os, os.path as osp
```
mstriemer: I'd rather see this as

```py
import os
import os.path
```

But if you'd rather use `osp` then put it on its own line I'd say. I'm guessing `os.path` makes some of these lines more than 80 characters but since we aren't PEP8ing I'd rather see a longer line than a non-standard usage of `os.path`.

mstriemer: If you want to make it shorter you could do:

```py
relative = lambda *paths: os.path.join(path, *paths)

File: extras/update_hashes.py [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+import subprocess
+
+path = osp.split(__file__)[0]
+target = "jslibs"
+
+subprocess.call(["python", osp.join(path, "jslibfetcher.py"), target])
+subprocess.call(["python", osp.join(path, "build_hashes.py"), target])
+
+for file in glob.glob(osp.join(path, target, "hashes-*.txt")):
+	os.rename(file,
```
mstriemer: You've got some hard tabs here.

extras/update_hashes.py
```diff
@@ -0,0 +1,15 @@
...
+	  osp.join(path, osp.pardir, "validator", "testcases", osp.split(file)[1]))
+
+print "\nThird party library hashes updated."
```
mstriemer: And there should be a newline at the end of the file.

extras/update_hashes.py
```diff
@@ -0,0 +1,15 @@
...
+
+path = osp.split(__file__)[0]
+target = "jslibs"
+
+subprocess.call(["python", osp.join(path, "jslibfetcher.py"), target])
+subprocess.call(["python", osp.join(path, "build_hashes.py"), target])
+
+for file in glob.glob(osp.join(path, target, "hashes-*.txt")):
+	os.rename(file,
+	  osp.join(path, osp.pardir, "validator", "testcases", osp.split(file)[1]))
```
mstriemer: It looks like you can do [`os.path.basename(file)`](https://docs.python.org/2/library/os.path.html#os.path.basename) instead of `os.path.split(file)[1]`.

extras/build_hashes.py
```diff
@@ -0,0 +1,17 @@
+import sys
+import os
+import os.path as pth
```
mstriemer: I see you didn't change this but I think using `os.path` would be nicer.

extras/build_hashes.py
```diff
@@ -0,0 +1,17 @@
...
+import hashlib
+
+libs_folder_root = pth.join(pth.abspath(pth.split(__file__)[0]), sys.argv[1])
```
mstriemer: `os.path.split(__file__)[0]` could be `os.path.dirname(__file__)`.

extras/build_hashes.py
mstriemer: Apparently you don't need to import `os.path` if you import `os`, I thought there was some weird reason you had to.

extras/build_hashes.py
```diff
@@ -0,0 +1,17 @@
...
+for entry in os.listdir(libs_folder_root):
+    entry_path = os.path.join(libs_folder_root, entry)
+    if os.path.isdir(entry_path):
+        rule_folder = entry_path
```
mstriemer: Looks like this can go away and use `entry_path` everywhere.

wagnerand:
Background: https://github.com/mozilla/addons-linter/issues/598

extras/jslibfetcher.py
```diff
@@ -15,339 +21,386 @@ def process(url, destination):
...
-    "1.9.0",
-    "1.9.1",
-    "1.9.2",
-    "1.9.3",
-]
+        "allowed", "PersonasInteractive_bootstrap.js")
+
+
+BACKBONE_VERSIONS = {
+    "ALLOWED" : [
```
mstriemer: It might be a good idea to put these rule names in constants so that they don't get typod. We also don't normally put a space before the `:` in dicts.

```py
ALLOWED = "ALLOWED"
WARNING = "WARNING"

BACKBONE_VERSIONS = {
    ALLOWED: [
        "1.0.0",
        ...
    ]
}

wagnerand: In a follow-up, I'd really like to refactor those dicts into a separate file, probably JSON.

mstriemer:
Looks good!

r+wc

wagnerand:
Thanks so much for this fast review!

PR #417, opened 2016-04-15
Cleanup JS third-party lib fetching

wagnerand:
This does a bunch of things:

* Add proper output
* All files are fetched from HTTPS and the official maintainer (if available)
* Catch file download errors (we had a couple of 404 hashes in the past and no one noticed)
* Catch duplicate hashes
* Remove very old or discontinued versions/libraries
* Add new jQuery versions

r? @mstriemer


extras/build_hashes.py
```diff
@@ -11,5 +14,11 @@
...
                 path = os.path.join(entry_path, filename)
                 hash = hashlib.sha256(open(path).read()).hexdigest()
-                print path, hash
-                output.write('{} {}\n'.format(hash, os.path.basename(path)))
+                if hash not in hashes:
+                    print path, hash
+                    output.write('{} {}\n'.format(hash, os.path.basename(path)))
+                    hashes[hash] = path
+                else:
+                    print "Error: {} and {} have the same hash ({})...Exiting.".format(path, hashes[hash], hash)
```
mstriemer: Does this not normally happen? How do you resolve a case where they hash the same or do we never include multiple sources for the same file? ie jQuery from GitHub and from CDN JS?

wagnerand: This should not happen. This is a fallback where the URL we retrieve is not the actual script file but e.g. a 404 page, but the status code was not 404. It's just a safety net to point out that something fishy is going on. The case is resolved by examining both URLs and determining which one is wrong.

extras/jslibfetcher.py
```diff
@@ -230,308 +234,162 @@ def get_pattern(prefix, url_pattern, library):
...
+
+    # Underscore
+    get_pattern("underscore",
+                "https://raw.github.com/documentcloud/underscore/%s/underscore.js",
+                UNDERSCORE_VERSIONS)
+    get_pattern("underscore",
+                "https://raw.github.com/documentcloud/underscore/%s/underscore-min.js",
+                UNDERSCORE_VERSIONS)
+
+    CRYPTO_FILES = {
```
mstriemer: Should this move up with the other libraries?

extras/jslibfetcher.py
```diff
@@ -230,308 +234,162 @@ def get_pattern(prefix, url_pattern, library):
...
+
+    get_pattern(
+        "crypto_js", "https://crypto-js.googlecode.com/svn/tags/3.1.2/src/%s.js",
+        CRYPTO_FILES)
+
+
+print "Downloading third-party library files..."
+get_patterns()
+if download_errors:
+    for (url, code, reason) in download_errors:
```
mstriemer: I don't think these parens are required.

mstriemer:
This looks good. Definitely a good improvement, thanks!

One thing that I find unfortunate about how this was originally done is that it has the different library versions in distinct constants. If they were instead all in one big dict then the pattern could be part of the library data and the `get_pattern` calls could go away. `get_patterns` would be able to be replaced with `errors = download_all_files(KNOWN_LIBRARIES)`. I don't think this is necessarily something that has to change it just seems odd that it wasn't done that way originally to me.

r+wc

extras/jslibfetcher.py
wagnerand: This has been like that before, I did not change the order. I think it would make more sense to keep this close to the corresponding `get_pattern` because it does contain versions of one single library, but instead uses the a particular (=the latest) version for several files of a library collection.

wagnerand:
Thank you!

PR #421, opened 2016-04-20
Support correctly decoding BOM encoding for manifest.json.

EnTeQuAk:
Fixes mozilla/addons-server#2038


mstriemer:
It never occurred to me that this could be a thing. You'd think JSON libraries would handle this but it seems like eventually it happens in most languages. Crazy.

r+

EnTeQuAk:
yeah, especially for a generall purpose json loading library and since the RFC appears to be saying something like "there must be no BOM".

Thanks for the review!

PR #423, opened 2016-04-21
Warn for latest angularjs library only

wagnerand:


tofumatt:
Because the rest aren't allowed, they'll just fail, right? This means the latest one is allowed but warns users to upgrade?

wagnerand:
Self r+  👀

wagnerand:
Oh, github didn't update the page for me :)

Yes, there was a slight misunderstanding of what the actual policy should be. We should disallow all old versions, allow the most recent one (for the time being) but warn developers and editors about it.

tofumatt:
Cool, all good then (r+)

wagnerand:
Thank you.

PR #425, opened 2016-04-25
Fix Fx47 compat tests

wagnerand:
r? @EnTeQuAk

### DIFF
--- a/tests/compat/test_gecko40.py
+++ b/tests/compat/test_gecko40.py
@@ -46,13 +46,6 @@ class TestFX40Compat(CompatTestCase):
         self.assert_silent()
         self.assert_compat_warning(type_='warning')
 
-    def test_Application(self):
-        self.run_script_for_compat("""
-            Application.restart();
-        """)
-        self.assert_silent()
-        self.assert_compat_warning(type_='warning')
-
     def test_Dict_jsm(self):
         self.run_script_for_compat("""
             Components.utils.import("resource://gre/modules/Dict.jsm");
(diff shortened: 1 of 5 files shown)


tests/compat/test_gecko40.py
```diff
@@ -46,13 +46,6 @@ def test_extIApplication(self):
         self.assert_silent()
         self.assert_compat_warning(type_='warning')
 
-    def test_Application(self):
```
wagnerand: The validation currently does not support multiple checks for the same instance call. But as compat tests for 40 ran a long time ago and won't run again, we should be ok.

tests/compat/test_gecko47.py
```diff
@@ -89,7 +95,11 @@ def test_changed_return_type(self):
         assert self.compat_err.warnings[0]['compatibility_type'] == 'error'
         assert self.compat_err.warnings[0]['message'].startswith(message)
 
-        script = 'let slots = testModule.listSlots();'
+        script = '''
+            var modules = modules.QueryInterface(Components.interfaces.nsIPKCS11Module);
+            modules.listSlots();
+        '''
+        self.run_script_for_compat(script)
```
wagnerand: The original patch did not actually run the script.

EnTeQuAk: great catch, sadpanda

EnTeQuAk:
looks great, lgtm.

r+

wagnerand:
Thank you.

PR #435, opened 2016-06-04
Use correct filename(s) for binary warning

### LINKED ISSUE #411: Validator misflags png as binary when another binary file exists
https://addons.mozilla.org/en-US/developers/addon/flash-debugger/file/421089/validation

```
Flagged file extensions found.

Warning: Files whose names end with flagged extensions have been found in the add-on.

The extension of these files are flagged because they usually identify binary components. 
Please see http://addons.mozilla.org/developers/docs/policies/reviews#section-binary for 
more information on the binary content review process.

data/flashfirebug.swf
    data/html/images/themes/default/inspector/UIComponents/VolumeBar.png
```

The png file is not a binary file, nor should it be flagged.

wagnerand:
Fixes #411

### DIFF
--- a/tests/test_packagelayout.py
+++ b/tests/test_packagelayout.py
@@ -17,6 +17,8 @@ def test_blacklisted_files():
                    packagelayout.test_blacklisted_files,
                    True)
     assert err.metadata['contains_binary_extension']
+    assert any(warning['id'][1] == 'test_blacklisted_files'
+        and warning['file'] == 'omgitsadll.dll' for warning in err.warnings)
     assert not any(count for (key, count) in err.compat_summary.items())
 
     # Run the compatibility test on this, but it shouldn't fail or produce
--- a/validator/testcases/packagelayout.py
+++ b/validator/testcases/packagelayout.py
@@ -122,7 +122,7 @@ def test_blacklisted_files(err, xpi_package=None):
                 editors_only=True,
                 signing_help=HELP,
                 signing_severity='medium',
-                filename=name)
+                filename=', '.join(flagged_files))
 
 
 @decorator.register_test(tier=1)


EnTeQuAk:
I would love to see a minimal test verifying that the output is indeed what we expect since now `filename` is a comma separated list and cannot be resolved to a particular filename anymore.

Isn't the problem in #411 a different one? I mean, the error reported there is also referencing the actual description which uses `flagged_files` so somehow the PNG ends up there.

I don't see exactly why that should be the case (the code is relatively simple, if extension in blacklisted_extensions... and there is no png in there) but maybe a test reveals a bit more details.

wagnerand:
The problem is that in https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L65 we loop over all files to identify the ones we should flag. However handling of flagged files is outside the loop (https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L94) but we use the loop variable `name` in https://github.com/mozilla/amo-validator/blob/master/validator/testcases/packagelayout.py#L125 for the filename to flag, which happens to be the last checked file from the loop.

I considered moving the warning into the loop, but it's generally more useful if we have one warning and just list _n_ occurrences instead of having _n_ warnings with one occurrence each.

EnTeQuAk:
fair enough, let's try this and see if there's more hiding.

But can you add a very simple test at least somewhere in `tests/test_packagelayout.py`? There are already a few and I believe it should be reasonably easy to add a simple one to verify it's working as expected.

wagnerand:
@EnTeQuAk I adapted an existing test. Mind having another look?

tests/test_packagelayout.py
```diff
@@ -17,6 +17,8 @@ def test_blacklisted_files():
                    packagelayout.test_blacklisted_files,
                    True)
     assert err.metadata['contains_binary_extension']
+    any(warning['id'][1] == 'test_blacklisted_files'
```
EnTeQuAk: guess there's an `assert` missing?

wagnerand: Ugh, the copy-and-paste devil. Thanks.

EnTeQuAk:
r+ once travis passes

wagnerand:
Thank you!

PR #436, opened 2016-06-07
Firefox 48 compatibility tests

EnTeQuAk:


validator/testcases/regex.py
```diff
@@ -426,6 +426,42 @@ def js_tests(self):
...
 
+@register_generator
+class Gecko48RegexTests(CompatRegexTestHelper):
+    """Regex tests for Firefox 48 updates."""
+
+    VERSION = FX48_DEFINITION
+
+    def js_tests(self):
+        yield self.get_test(
+            r'\bProxy\.(create|createFunction)\b',
```
wagnerand: I am afraid this is bound to create a huge amount of false positives.

wagnerand:
r+

validator/testcases/regex.py
EnTeQuAk: I fixed this in 3dcebf8 to not use a regex which should produce much less false positives.

wagnerand:
r+ with ❤️

EnTeQuAk:
Awww :smiley:

PR #437, opened 2016-06-10
Don't warn for e10s if no install.rdf is present, like in .jar files

diox:
We re-enter `test_inner_package()` if the add-on contains a .jar file, which could cause superfluous warnings. We want to warn twice for add-ons bundling other add-ons if both are incompatible, but we don't want to warn twice if the subpackage is just a .jar file.

Fix mozilla/addons-server#2881


EnTeQuAk:
r+

> without mocks

<3

PR #443, opened 2016-06-27
Don't show e10s compatibilty warning for non-extensions that have an install.rdf

diox:
Fix mozilla/addons-server#2998


validator/submain.py
```diff
@@ -386,6 +386,7 @@ def test_inner_package(err, xpi_package, for_appversions=None):
     if (err.get_resource('has_install_rdf') and
             not err.get_resource('has_manifest_json') and
             not err.get_resource('is_multiprocess_compatible') and
+            err.detected_type == PACKAGE_EXTENSION and
```
eviljeff: should we include `PACKAGE_MULTI` and `PACKAGE_SUBPACKAGE` ? (not sure what these are exactly)

diox: `PACKAGE_MULTI` value is `1` just like `PACKAGE_EXTENSION` so it's already covered. We probably do need to include`PACKAGE_SUBPACKAGE`, though.

EnTeQuAk: though, explicit is better than implicit regarding the inclusion of both `PACKAGE_MULTI` and `PACKAGE_EXTENSION` so I'm actually curious why we have separate constants if we don't use them :-/ But I gues that's "legacy".

No idea what exactly a subpackage is though :disappointed:

diox: `PACKAGE_SUBPACKAGE` is for extensions that bundle other extensions, AFAIK

diox:
re-r? I'm not quite sure how to test with subpackages (the whole thing appears to be untested accross the repos) but I don't think it should be a problem.

EnTeQuAk:
r+

PR #450, opened 2016-07-29
Add appversion to Proxy compat test

wagnerand:
We forgot to add this, so this test currently runs for every appversion, which is not what we want.


EnTeQuAk:
r+

wagnerand:
Thank you!

Yeah I will tag on Monday, will have to check whether they are any new third party libs to add.

PR #451, opened 2016-08-01
Update third party libraries

wagnerand:


wagnerand:
This depends on #452.

validator/testcases/hashes-allowed.txt
```diff
@@ -110,7 +126,6 @@ e29720f3758289a539385d5e833158e726f6218f0c57eb7c0055b8a539fe3dcb dojo.1.9.6.dojo
 049bf422aa804e91e16d38921db5408a6e8f60253e7e7b876da17139fb7ffb9b dojo.1.9.8.dojo.js.uncompressed.js
 f7dced86cb81108ca8312a6c049f91d9cef2bc3f331a55179e11729c0b5e8f14 dojo.1.9.9.dojo.js
 cf07ff8ad6b598024ca8ad3535bfb0d5c443d203d160771ce4c37150708457f4 dojo.1.9.9.dojo.js.uncompressed.js
-77988cda4c800c18f58ade9de769083f2db329a46269cb5bb942a3ae1c4274cb dompurify.0.6.7.purify.js
```
diox: Where did it come from... Were the versions in `jslibfetcher.py` updated previously without running the script?

diox:
r+wc

validator/testcases/hashes-allowed.txt
wagnerand: The code currently takes all files in `extras/jslibs`. If you removed an entry in `jslibfetcher.py` you need to remember to either prune the whole directory or at least remove the file as well. I might have forgotten that during the last cleanup. It's not a big issue though.

My goal is to have the same hash list for dispensary and amo-validator which is still WiP, but we're now closer than ever.

wagnerand:
Thank you!

PR #457, opened 2016-08-05
Remove unused files

wagnerand:
This isn't used these days, not sure what it was ever needed for.


diox:
lolwut r+

wagnerand:
@diox sorry I just added another unused (and broken) file. Lots of cleanups...

functional_tests/test_validator.py
```diff
@@ -18,8 +18,6 @@ def _validator(file_path, for_appversions=None, overrides=None):
     import validator
     import validator.constants
     js = os.environ.get('SPIDERMONKEY_INSTALLATION', 'js')
-    scripting.SPIDERMONKEY_INSTALLATION = js
-    validator.constants.SPIDERMONKEY_INSTALLATION = js
```
diox: But then you don't need the line above either, right ?

diox: err actually scratch that

diox: so `validate()` does the right thing when you pass a custom `spidermonkey`, and that's why you're removing those references to `SPIDERMONKEY_INSTALLATION` ?

wagnerand: `SPIDERMONKEY_INSTALLATION` isn't used in validator code at all. So it made no sense to me to set it.

wagnerand:
Thank you!

PR #469, opened 2016-09-14
Add React hashes

diegocr:
React hashes for versions 0.14.8, 15.0.0, and 15.3.0

An additional improvement for the Validator could be that if React is found it also checks for `dangerouslySetInnerHTML` usages and related.


diegocr:
@wagnerand ping ;)

wagnerand:
React hashes are already in https://github.com/mozilla/amo-validator/blob/master/validator/testcases/hashes-allowed.txt#L276-L364

Those files are the main files offered by React developers on their website. I'm not sure we should support other files as well. We don't do that for jQuery and such either.

PR #470, opened 2016-09-30
Add JS 'forbid' list and add angular and old jQuery

wagnerand:
We no longer allow angular 1.x and old jQuery versions, so we should error during submission instead of letting it pass, have the add-on wait for days in the queues and finally getting manually rejected by a reviewer.


wagnerand:
Adding @kewisch so he is aware.

Issue #476, opened 2016-10-14
[fx51] Show compatibility error if an add-on uses onButtonClick


jvillalobos:
As part of [bug 1167575](https://bugzilla.mozilla.org/show_bug.cgi?id=1167575), the behavior of `onButtonClick` changed. All instances of this function should be flagged.

This is a compatibility error, so the add-on shouldn't be upgraded in this case. It should also appear as a warning in regular validations.

Message:

The function onButtonClick is now asynchronous. For more information, please refer to [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1167575).

Affected add-on: https://addons.mozilla.org/addon/imacros-for-firefox/

Release timing: Firefox 51 will be released on January 24th.

Issue #475, opened 2016-10-14
[fx51] Show compatibility warning if an add-on uses setAndFetchFaviconForPage or replaceFaviconDataFromDataURL


jvillalobos:
As part of [bug 1227289](https://bugzilla.mozilla.org/show_bug.cgi?id=1227289), the functions `setAndFetchFaviconForPage` and `replaceFaviconDataFromDataURL` have changed their arguments and default behavior. All instances of these should be flagged.

This is a compatibility warning, so the add-on should still be upgraded in this case. It should also appear as a warning in regular validations.

Message:

setAndFetchFaviconForPage and replaceFaviconDataFromDataURL now default to using a null principal for security reasons. An appropriate principal should be passed if different behavior is required. For more information, please refer to [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1227289).

Affected add-on: https://addons.mozilla.org/addon/xmarks-sync/

Release timing: Firefox 51 will be released on January 24th.

Issue #474, opened 2016-10-14
[fx51] Show compatibility error if an add-on uses mozVisibilityState or mozHidden


jvillalobos:
As part of [bug 812701](https://bugzilla.mozilla.org/show_bug.cgi?id=812701), the `mozVisibilityState` and `mozHidden` properties have been unprefixed. All instances of these should be flagged.

This is a compatibility error, so the add-on shouldn't be upgraded in this case. It should also appear as a warning in regular validations.

Message:

The mozVisibilityState and mozHidden properties are no longer prefixed, so you should use visibilityState and hidden instead. For more information, please refer to [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=812701).

Affected add-on: https://addons.mozilla.org/addon/scrollbar-search-highlighter/

Release timing: Firefox 51 will be released on January 24th.

PR #479, opened 2016-10-24
Fx51 compat tests

### LINKED ISSUE #473: [fx51] Show compatibility error if an add-on uses BrowserOpenNewTabOrWindow
As part of [bug 528005](https://bugzilla.mozilla.org/show_bug.cgi?id=528005), the `BrowserOpenNewTabOrWindow` function has been removed. All instances of this function should be flagged.

This is a compatibility error, so the add-on shouldn't be upgraded in this case. It should also appear as a warning in regular validations.

Message:

The function BrowserOpenNewTabOrWindow has been removed. You can use BrowserOpenTab instead, but its behavior is not identical. For more information, please refer to [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=528005).

Affected add-on: https://addons.mozilla.org/addon/tab-tree/

Release timing: Firefox 51 will be released on January 24th.

wagnerand:
Fixes #473, #474, #475, #476

### DIFF
--- a/tests/compat/test_gecko51.py
+++ b/tests/compat/test_gecko51.py
@@ -1,0 +1,148 @@
+import pytest
+
+from helper import CompatTestCase
+from validator.compat import FX51_DEFINITION
+
+
+class TestFX51Compat(CompatTestCase):
+    """Test that compatibility tests for Gecko 50 are properly executed."""
+
+    VERSION = FX51_DEFINITION
+
+    def test_mozIAsyncFavicons_setAndFetchFaviconForPage_changed(self):
+        """https://github.com/mozilla/amo-validator/issues/475"""
+        expected = (
+            'The methods setAndFetchFaviconForPage and '
+            'replaceFaviconDataFromDataURL now default to using a null '
+            'principal for security reasons. An appropriate principal should '
+            'be passed if different behavior is required.')
+
+        script = '''
+            var ioService = Components.classes["@mozilla.org/network/io-service;1"].
+                getService(Components.interfaces.nsIIOService);
+            var pageUri = ioService.newURI('http://mozilla.org', 'utf-8', 'mozilla.org');
+            var faviconUri = ioService.newURI('http://mozilla.org/fav.ico', 'utf-8', 'mozilla.org');
+
+            var faviconService = Components.classes["@mozilla.org/browser/favicon-service;1"].
+                getService(Components.interfaces.nsIFaviconService).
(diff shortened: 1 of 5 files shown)


tests/compat/test_gecko51.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_mozHidden_unprefixed(self):
+        """https://github.com/mozilla/amo-validator/issues/474"""
+        expected = (
+            'The mozVisibilityState and mozHidden properties are no longer '
+            'prefixed.')
+
+        script = '''
+            var x = document.createElement('div');
+            div.mozHidden = true;
```
EnTeQuAk: even though it probably doesn't make a difference for the test but `div` should be `x` in this example or am I reading this wrong?

tests/compat/test_gecko51.py
```diff
@@ -0,0 +1,148 @@
...
+
+        warning = [
+            msg for msg in self.compat_err.warnings
+            if msg['message'].startswith(expected)]
+        assert warning
+        assert warning[0]['compatibility_type'] == 'error'
+
+    def test_onButtonClick_changed(self):
+        """https://github.com/mozilla/amo-validator/issues/476"""
+        expected = ('The function onButtonClick is now asynchronous.')
```
EnTeQuAk: you don't need the parens here.

tests/compat/test_gecko51.py
```diff
@@ -0,0 +1,148 @@
...
+
+        script = '''
+            var f = function() { return true };
+            var x = document.createElement('div');
+            div.onButtonClick(f);
```
EnTeQuAk: Same here, this should be `x` instead of `div` I think.

validator/testcases/regex.py
```diff
@@ -478,6 +478,41 @@ def tests(self):
         )
 
 
+@register_generator
+class Gecko51RegexTests(CompatRegexTestHelper):
+    """Regex tests for Firefox 50 updates."""
```
EnTeQuAk: Should be "Firefox 51"

validator/testcases/regex.py
```diff
@@ -478,6 +478,41 @@ def tests(self):
...
+
+    VERSION = FX51_DEFINITION
+
+    def tests(self):
+        yield self.get_test_bug(
```
EnTeQuAk: Lol, I didn't know that helper existed. Nice!

PR #482, opened 2016-11-04
remove blacklist

andymckay:
When we make a new version of this and bump it in AMO, we'll need to alter AMO to use the new imports, not the existing ones.


validator/testcases/content.py
```diff
@@ -157,7 +157,7 @@ def test_packed_packages(err, xpi_package=None):
                 err.notice(
                     err_id=('testcases_content',
                             'test_packed_packages',
-                            'blacklisted_js_library'),
+                            'denied_js_library'),
```
wagnerand: ⚠️ We explicitly check for this exact string in https://github.com/mozilla/addons-server/blob/master/static/js/zamboni/files.js#L524 to update the validation report and file viewer.

wagnerand: Ironically, we use this to indicate a JS library we *allow*, instead of *deny*.

andymckay: Well that's just slightly confusing. Thanks I'll go back and actually read this instead of doing a quick search and replace.

Issue #488, opened 2016-11-23
Identify hybrid add-ons


wagnerand:
Following: https://github.com/mozilla/addons-server/issues/2875

The validator should raise an *information* (displayed to reviewers only) with the following details:

Title: This add-on is a hybrid add-on.
Description: The folder `webextension` contains an embedded webextension. Handle validation messages with care. For more information, refer to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Embedded_WebExtensions

wagnerand:
I am wondering whether including the `webextension` folder in the scan does more good or more harm.

@kewisch @jvillalobos What do you think?

rpl:
@wagnerand I just tried to run the existent amo-validator rules on the xpi files generated from the two example add-ons of the mdn/webextensions-examples repo:

- hybrid bootstrapped add-on: https://github.com/mdn/webextensions-examples/tree/master/embedded-webextension-bootstrapped/step1-hybrid-addon
- hybrid sdk add-on: https://github.com/mdn/webextensions-examples/tree/master/embedded-webextension-sdk/step1-hybrid-addon

By running the xpi files created from the above two add-on directories we can immediatelly notice the first change that we have to apply to the amo-validator:

```
Summary:
------------------------------
Detected type: Extension/Multi-Extension
------------------------------
Test failed! Errors:

....
Notice: Unrecognized element in install.rdf
```

which should instead detect the new `hasEmbeddedWebExtension` and show the notice message that @wagnerand suggests in this issue summary.

it is the first time that I look into the amo-validator sources, but it seems that we can achieve this by applying something similar to the following changes:

```diff
diff --git a/validator/testcases/installrdf.py b/validator/testcases/installrdf.py
index bdc71f7..70ca997 100644
--- a/validator/testcases/installrdf.py
+++ b/validator/testcases/installrdf.py
@@ -44,6 +44,7 @@ def _test_rdf(err, install):
                       'homepageURL',
                       'creator',
                       'multiprocessCompatible',
+                      'hasEmbeddedWebExtension',
                       'optionsType',
                       'type',
                       'updateKey',
@@ -93,6 +94,21 @@ def _test_rdf(err, install):
         elif predicate == 'multiprocessCompatible':
             err.save_resource('is_multiprocess_compatible', value == 'true')

+        elif predicate == 'hasEmbeddedWebExtension':
+            err.notice(('testcases_installrdf',
+                        '_test_rdf',
[...]

PR #489, opened 2016-11-23
Detect hybrid add-ons xpis and print a specific notice message.

rpl:
This PR introduces the changes needed to make the amo-validator able to detect an hybrid add-on and print a more detailed notice message that can be helpful to the reviewers (related to issue #488).

- [x] detect the new `hasEmbeddedWebExtension` install.rdf property
- [x] associate a detailed notice message to the addon when the above property has been found
- [x] add tests related to the above changes


rpl:
I'm going to look into the testing part of this change.

validator/testcases/installrdf.py
```diff
@@ -93,6 +94,21 @@ def _test_rdf(err, install):
         elif predicate == 'multiprocessCompatible':
             err.save_resource('is_multiprocess_compatible', value == 'true')
 
+        elif predicate == 'hasEmbeddedWebExtension':
+            err.notice(('testcases_installrdf',
+                        '_test_rdf',
+                        'hasEmbeddedWebExtension'),
+                       'This add-on is an hybrid add-on',
```
wagnerand: Actually, would you mind changing this to `This add-on contains an embedded webextension.`, please? I realize it makes more sense to stay consistent with our terminology and we use "embedded webextensions" on MDN.

validator/testcases/installrdf.py
```diff
@@ -93,6 +94,21 @@ def _test_rdf(err, install):
...
+                       "The element \"%s\" was found in the add-on's "
```
wagnerand: I think we can remove that sentence.

validator/testcases/installrdf.py
rpl: @wagnerand yeah, it is definitely better to keep it consistent to the terminology used on MDN.

updated patch is coming

wagnerand:
This lgtm.

Let's see if @EnTeQuAk agrees 😄

tests/test_validate.py
```diff
@@ -350,3 +352,39 @@ def test_validate_dictionnary_no_multiprocess_compatible_warning():
...
+    assert data['compatibility_summary']['errors'] == 0
+    assert data['compatibility_summary']['notices'] == 0
+    assert data['compatibility_summary']['warnings'] == 0
+    assert data['detected_type'] == 'extension'
+
+    assert len(data['messages']) == 1
+    assert data['messages'][0]['file'] == u'install.rdf'
+    assert data['messages'][0]['message'] == u'This add-on contains an embedded webextension'
+
+    expectedHybridDescriptionRegExp = re.compile (
```
EnTeQuAk: the naming doesn't follow pep08, should be something like `expected_hybrid_description_regex` or something like that.

tests/test_validate.py
```diff
@@ -350,3 +352,39 @@ def test_validate_dictionnary_no_multiprocess_compatible_warning():
...
+        '^The embedded webextension is in the folder "webextension"\. .*'
+        'For more information, refer to https://developer.mozilla.org/.*Embedded_WebExtensions$'
+    )
+    assert expectedHybridDescriptionRegExp.match(data['messages'][0]['description'])
```
EnTeQuAk: This could be simplified to just do something like `assert 'The embedded webextension is in the folder "webextension"' in data['messages'][0]['description']`

tests/test_validate.py
```diff
@@ -350,3 +352,39 @@ def test_validate_dictionnary_no_multiprocess_compatible_warning():
...
+    assert data['message_tree']
+    assert data['signing_summary']['high'] == 0
+    assert data['signing_summary']['medium'] == 0
+    assert data['signing_summary']['low'] == 0
+    assert data['signing_summary']['trivial'] == 0
+    assert data['metadata']
+    assert data['metadata']['name'] == u'name_value'
+    assert data['metadata']['version'] == u'1.2.3.4'
+    assert data['metadata']['listed'] is True
+    assert data['metadata']['id'] == u'an-hybrid-addon@mozilla.com'
```
EnTeQuAk: The test doesn't test the `has_embedded_webextension` flag that gets set. In theory it should be in `data['resources']` or something like that, mind adding that?

rpl: @EnTeQuAk :+1: I was looking for how to do it, that field is not in the properties serialized in the JSON format, I took a deeper look and I found how to do it:
I added an additional step that retrieve the "error bundle" instead of the json result, and I've been able to add an assertion for it :tada:

rpl:
@EnTeQuAk Thanks a lot, I had the feeling that the test needed some tweak to better follow the usual python coding style conventions!

I just pushed the updated version which should include the suggested changes and some additional minor tweaks (e.g. the name of the test function)

tests/test_validate.py
EnTeQuAk: Yeay, awesome! I wasn't sure either where it'll be serialized, thanks for digging deepter!

PR #495, opened 2017-01-12
Add 'is_extension' to validator metadata

diox:
Another bit needed for legacy submissions restrictions. I thought `type` was already exposed in `metadata`, but it's not, so I'm just adding this small bit.


validator/submain.py
```diff
@@ -69,6 +69,9 @@ def timeout_handler(signum, frame):
             package = open(path, 'rb')
             test_package(err, package, path, expectation, for_appversions)
 
+        if err.detected_type == PACKAGE_EXTENSION:
+            err.metadata['is_extension'] = True
+
```
wagnerand: why wouldn't we set `err.metadata['is_extension'] = False` in the _else_ case?

validator/submain.py
```diff
@@ -69,6 +69,9 @@ def timeout_handler(signum, frame):
             package = open(path, 'rb')
             test_package(err, package, path, expectation, for_appversions)
 
+        if err.detected_type == PACKAGE_EXTENSION:
+            err.metadata['is_extension'] = True
```
EnTeQuAk: I wonder if, given the name this should always be `False` or `True` but never not given. Although we do have a few other keys that behave this way I think. Still, can we make this explicitly set?

EnTeQuAk:
And there are tests failing :)

diox:
Interesting that test doesn't fail for me locally... sigh.

validator/submain.py
diox: My reasoning was that:
1) I don't know all possible weird edge cases in validator that might cause the code to use a different path and avoid that function altogether
2) The existing code already behaves this way
3) In addons-server, where we have things coming from this validator but also from the linter, we will have to assume the key could be absent anyway

EnTeQuAk: The linter bit is a good reasoning. I don't think there's any other code-path, `submain` is pretty common to be called, imho it's always called but I could easily be wrong.

PR #499, opened 2017-01-26
Revert Add public warning about using ctypes

wagnerand:
Reverting because by the time you'd see the public warning, you aren't part of the target audience for that warning anymore because after the validation ran, the add-on has already been created and subsequent submissions are just updates that aren't affected.


muffinresearch:
Does this mean the subsequent package update post this landing needs to be cherry-picked for release to prevent the warning going out today?

PR #520, opened 2017-03-20
Fix Fx53 compat tests regex's

wagnerand:
@EnTeQuAk I tried to run the Fx53 compat tests today and they keep timing out. After some digging, I am fairly confident that either or both of these regex's are utterly slow and cause the validation to hit the time limit for some add-ons.

### DIFF
--- a/validator/testcases/regex.py
+++ b/validator/testcases/regex.py
@@ -537,7 +537,7 @@ class Gecko53RegexTests(CompatRegexTestHelper):
     def tests(self):
         yield self.get_test_bug(
             1321556,
-            r'\burlbarBindings.xml\#splitmenu\b',
+            r'\burlbarBindings\.xml#splitmenu\b',
             'The splitmenu element has been removed.',
             'The splitmenu element has been removed.',
             log_function=self.err.warning,
@@ -545,7 +545,7 @@ class Gecko53RegexTests(CompatRegexTestHelper):
 
         yield self.get_test_bug(
             1331296,
-            r'\b[^-]*-moz-calc\b',
+            r'-moz-calc\b',
             'The -moz-calc function has been removed.',
             'You can use the equivalent calc instead',
             log_function=self.err.warning,


validator/testcases/regex.py
```diff
@@ -537,15 +537,15 @@ class Gecko53RegexTests(CompatRegexTestHelper):
...
+            r'\burlbarBindings\.xml#splitmenu\b',
             'The splitmenu element has been removed.',
             'The splitmenu element has been removed.',
             log_function=self.err.warning,
             compat_type='error')
 
         yield self.get_test_bug(
             1331296,
-            r'\b[^-]*-moz-calc\b',
+            r'-moz-calc\b',
```
wagnerand: This could cause some false-positives. We could potentially narrow it down to `\b[^\w]+-moz-calc\b` but I assume it's as slow as the previous one.

validator/testcases/regex.py
```diff
@@ -537,15 +537,15 @@ class Gecko53RegexTests(CompatRegexTestHelper):
     def tests(self):
         yield self.get_test_bug(
             1321556,
-            r'\burlbarBindings.xml\#splitmenu\b',
+            r'\burlbarBindings\.xml#splitmenu\b',
```
EnTeQuAk: nice find, unrelated though but okay.

validator/testcases/regex.py
EnTeQuAk: I think I did something stupid here, `-` can be used as a "from to" (e.g `a-zA-Z`) and apparently I didn't escape it properly. I wonder if that already helps, there was a reason I chose this regex since a few others (e.g `\b-moz-calc\b`) didn't properly match. Generally I'm 👍 for your approach but I think the actual problem can be fixed via `\b[^\-]*-moz-calc\b`

wagnerand: Usually, if `-` is the last character in square brackets, it does not need to be escaped. I think the `*` made it slow, as the match can be quite large and thus take some time to compute.

PR #525, opened 2017-04-10
Implement Firefox 54 tests.

EnTeQuAk:


tests/compat/test_gecko54.py
```diff
@@ -0,0 +1,37 @@
...
+            'FormHistory.jsm instead.')
+
+        code = '''
+            Components.classes["@mozilla.org/satchel/form-history;1"].getService(Components.interfaces.nsIFormHistory2)
+        '''
+
+        self.run_script_for_compat(code)
+
+        assert not self.compat_err.notices
+        assert not self.compat_err.errors
```
wagnerand: Can we add a `assert self.warnings ...` here to make sure it shows up in normal validations as a warning?

tests/compat/test_gecko54.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_moz_appearance(self):
+        """https://github.com/mozilla/amo-validator/issues/524"""
+        expected = '-moz-appearance can only be used in chrome stylesheets.'
+
+        invalid = '-moz-border-radius: 0px; -moz-appearance:textfield;'
+
+        self.run_regex_for_compat(invalid, is_js=False)
+        assert not self.compat_err.notices
+        assert not self.compat_err.errors
```
wagnerand: Same here, please add a check to make sure a normal warning is raised.

validator/testcases/regex.py
```diff
@@ -567,6 +568,20 @@ def tests(self):
...
+@register_generator
+class Gecko54RegexTests(CompatRegexTestHelper):
+    VERSION = FX54_DEFINITION
+
+    def tests(self):
+        yield self.get_test_bug(
+            1333482,
+            r'-moz-appearance\b',
+            '-moz-appearance can only be used in chrome stylesheets.',
+            'All other users will be ignored.',
```
wagnerand: I don't understand what this sentence is trying to say. Do you mean `uses`? Can we rephrase to `Any other usage ...` or something like that?

tests/compat/test_gecko54.py
EnTeQuAk: absolutely, I'll add that.

EnTeQuAk: Actually… that's what `self.assert_compat_error('warning', expected)` does, it makes sure it's a compat error and a `warning` otherwise :)

validator/testcases/regex.py
EnTeQuAk: Fair point, I honestly only copied from the issue. I feel like it should have been `All other uses will be ignored` which makes sense, let's stick to that?

tests/compat/test_gecko54.py
wagnerand: Cool beans! We rock!

validator/testcases/regex.py
wagnerand: Ah. Yeah, wfm.

PR #526, opened 2017-04-10
Update JS third party libs

wagnerand:


extras/jslibfetcher.py
```diff
@@ -343,7 +365,6 @@
 
 REACT_VERSIONS = {
     ALLOWED: [
-        "0.10.0",
```
wagnerand: That version is no longer available on the CDN.

PR #527, opened 2017-04-13
Bug 1355891: Simplify Marionette prefs check to catch all Marionette …

AutomatedTester:
…prefs.

### DIFF
--- a/validator/testcases/javascript/predefinedentities.py
+++ b/validator/testcases/javascript/predefinedentities.py
@@ -169,9 +169,7 @@ BANNED_PREF_BRANCHES = (
     (u'nglayout.debug.disable_xul_cache', None),
 
     # Marionette
-    (u'marionette.force-local', MARIONETTE_MESSAGE),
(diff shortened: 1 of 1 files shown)


EnTeQuAk:
Does this catch anything in particular that it didn't before? If so, can we add a test for it?

AutomatedTester:
@EnTeQuAk Test is already there in https://github.com/mozilla/amo-validator/blob/master/tests/test_regex.py#L16-L26

The prefs in the validator are no longer used and so simplifying to just the top level

wagnerand:
If they are no longer used, why not just remove them?

AutomatedTester:
@wagnerand because the prefs in the validator are not used. There are new ones hence this pull request.

@muffinresearch can we get this landed?

muffinresearch:
@wagnerand would you have time to land this and cut a release, and update addons-server to use it? Or alternatively if you can point me to docs on the package release process I can get it done.

PR #528, opened 2017-04-13
Add webextension-polyfill lib

wagnerand:
Although this will probably be rarely used (the only scenario I could imagine are embedded webextensions), it doesn't hurt.


wagnerand:
<@cgrebs> TheOne: I somehow can't login right now on my mobile, so r+ here to all of em :)

PR #558, opened 2017-07-21
[fx56] Show compatibility error if an add-on uses `showModalDialog`

### LINKED ISSUE #557: [fx56] Show compatibility error if an add-on uses `showModalDialog`
As part of [bug 981796](https://bugzilla.mozilla.org/show_bug.cgi?id=981796), `window.showModalDialog` has been removed. All instances of this function should be flagged.

This is a compatibility error, so the add-on shouldn't be upgraded in this case. It should also appear as a warning in regular validations.

Message:

The showModalDialog function has been removed. See [this bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=981796) for more information.

Affected add-on: https://addons.mozilla.org/addon/imtranslator/

EnTeQuAk:
Fixes #557


validator/testcases/regex.py
```diff
@@ -582,6 +582,23 @@ def tests(self):
...
+@register_generator
+class Gecko56RegexTests(CompatRegexTestHelper):
+    """Regex tests for Firefox 56 updates."""
+
+    VERSION = FX56_DEFINITION
+
+    def tests(self):
+        yield self.get_test_bug(
+            981796,
+            r'\bshowModalDialog\b',
```
EnTeQuAk: I tried to make this less prone to false positives with proper tree traversal but somehow didn't manage to get it being raised properly… I figured, we had regex tests for other functions like `onButtonClick` too so this might work.

PR #559, opened 2017-07-21
Add test to make sure we don't raise a false positive for createElementNS.

### LINKED ISSUE #368: False positive: createElement(NS)
The code `doc.createElementNS( nsXUL, 'panelview' )` causes false positive...

The warning about `createElement` is a bug in the validator since the `<script>` tag is never used, and since `nsXUL` is a constant which can't be changed.

An example here: https://github.com/Centril/firefox-line/blob/master/src/search-engine.js#L206

EnTeQuAk:
Fixes #368


wagnerand:
@EnTeQuAk could you add another test for `x.createElementNS(nsXUL, 'script')` please?

EnTeQuAk:
Added in b10225414

PR #581, opened 2017-12-13
Fix failing unittests on Python 2.7.14

EnTeQuAk:
Fix failing unittests on python 2.7.14

The problem here is that `sax` closes any file-like object you're
passing it to and these days StringIO forces operations on closed files
to fail.

It's hard to find the actual commit that caused this change in Python
2.7.14 but explicitly reading the file and wrapping it in a new StringIO
object works.

This isn't ideal but RDF files should not be too large so that this
workaround does sound reasonable to me.

### DIFF
--- a/validator/rdf.py
+++ b/validator/rdf.py
@@ -79,18 +79,20 @@ class RDFParser(object):
         self.manifest = u'urn:mozilla:install-manifest'
         self.namespace = namespace or 'http://www.mozilla.org/2004/em-rdf'
 
-        if isinstance(data, types.StringTypes):
-            data = StringIO(data)  # Wrap data in a pseudo-file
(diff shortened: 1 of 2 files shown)


validator/rdf.py
```diff
@@ -79,18 +79,20 @@ def __init__(self, err, data, namespace=None):
...
         try:
             # Use an empty ContentHandler, we just want to make sure it parses.
-            parse(data, ContentHandler())
+            parse(StringIO(data), ContentHandler())
         except SAXParseException as ex:
             # Raise the SAX parse exceptions so we get some line info.
             raise RDFException(orig_exception=ex)
-        else:
-            # We consumed the file, start over from the beginning.
-            data.seek(0)
```
EnTeQuAk: `sax` closes the file, so that `data.seek(0)` fails now.

PR #618, opened 2019-01-23
force amo-validator to be executed as python2

eviljeff:
when addons-server is executing this in a python3 env it's defaulting to py3 which isn't working (because this package is py2 only).  So force the python to be v2.


EnTeQuAk:
New version incoming

